PR #6, opened 2018-04-18
feature: Chart options component for Data vizualiser DHIS2-3279

joakimia:
This component enables filters and customization of chart's within the new data visualizer.

### DIFF
--- a/src/ChartOptions/ChartOptions.js
+++ b/src/ChartOptions/ChartOptions.js
@@ -61,7 +61,6 @@ export class ChartOptions extends Component {
         },
     };
     handleChange = (content, value) => {
-        console.log(content, value);
         content === 'activeTab'
             ? this.setState({ activeTab: value })
             : this.setState({
@@ -103,7 +102,6 @@ export class ChartOptions extends Component {
                         className={classes.hideButton}
                         color={'primary'}
                         size={'small'}
-                        //TODO: Dispatch settings to Redux store (?)
                     >
                         {i18n.t('Hide')}
                     </Button>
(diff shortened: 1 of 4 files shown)


src/options-component/DataTab.js
```diff
@@ -0,0 +1,174 @@
...
+                onChange(key, event.target.value);
+            }} // TODO (?) use parseInt on number? onChange will always pass a string
+            type={i % 2 === 0 ? 'number' : 'string'} // 1st and 3rd TextField component are number specific
+            value={value}
+            key={i}
+        />
+    ));
+};
+
+const renderSelectFields = (classes, tabContent, onChange) => {
```
jenniferarnesen: This function is not used. Can it be removed?

src/options-component/DataTab.js
```diff
@@ -0,0 +1,174 @@
...
+    //render first two select components
+    // call			{renderTextFields(classes, textFields, onChange)} (?)
+    //render last select component
+};
+
+const DataTab = props => {
+    const { classes, tabContent, onChange } = props;
```
jenniferarnesen: You can write this in a more shorthand way on line 63 (and drop line 64):

const DataTab = ({ classes, tabContent, onChange }) = {

It's pretty clean as long as there aren't too many props. More than 4, then it might be better to do it as you had done.

src/options-component/DataTab.js
```diff
@@ -0,0 +1,174 @@
...
+    let textFields = Object.entries(tabContent).slice(5, 9); // slice the prop object, to render DataTab's textfield values
+    return (
+        <div className={classes.divBorder}>
+            <DataTabCheckBoxes tabContent={tabContent} onChange={onChange} />
+
+            <FormControl className={classes.formControl}>
+                <InputLabel className={classes.inputLabeltextSize}>
+                    {' '}
```
jenniferarnesen: avoid using empty spaces for padding

jenniferarnesen: I'm thinking that each tab should be smart enough and know which props its going to display.  So an internal data-properties object that contains all the props that will be displayed instead of the list as you have in index.js. In fact, I think maybe we should make use of react component state, for the moment, although that will most likely get replaced with redux.

PR #8, opened 2018-05-04
Use d2-ui favorites menu and open chart options in dialog

edoardo:


package.json
```diff
@@ -33,8 +35,11 @@
         "css-loader": "0.28.7",
         "d2": "^29.1.5",
         "d2-i18n": "^1.0.2",
+        "d2-i18n-extract": "^1.0.4",
```
jenniferarnesen: These 2 packages have been scoped to dhis2 now. For example:

@dhis2/d2-i18n-generate

PR #27, opened 2018-09-20
Connect options to store

edoardo:
Kinda big PR, but it should be quite straightforward to review.
Basically we wanted to have as much as possible reusable UI components for each option, which eventually will be used unchanged in other analytic apps (ie. EV has pretty much the same options as DV).
Also, I tried to keep the options info (ie. the name) in as fewer places as possible.
I figured there are 3 main option types, checkbox, text and select, so there are now 3 base components that implement those with all the store connect stuff, we could probably reduce some duplication even more there...
All the options use the 3 base components and pass the relevant info in props.

The tabs component are much simplified this way, they basically just list and group the various options.
All styling is still to do, but I kept the style definitions from Joakim's version.
The tests need to be completely rewritten.


packages/app/src/options.js
```diff
@@ -1,46 +1,87 @@
...
     );
 };
+
+export const getOptionsFromVisualization = visualization => {
+    const optionsFromVisualization = {
+        ...getOptionsForUi(),
+        ...getPropsByKeys(visualization, Object.keys(options)),
+    };
+
+    if (
```
jenniferarnesen: To improve the readability of this code, what about something like this? (not tested)

```
const isNotDefault = prop => {
    return (
        optionsFromVisualization[prop] &&
        optionsFromVisualization[prop] !== options[prop].defaultValue
    );
};

optionsFromVisualization.baseLine =
    isNotDefault('baseLineLabel') || isNotDefault('baseLineValue');

optionsFromVisualization.targetLine =
    isNotDefault('targetLineLabel') || isNotDefault('targetLineValue');
```

PR #29, opened 2018-09-21
Allow onClick prop on UpdateButton

janhenrikoverland:


packages/app/src/components/MenuBar/UpdateButton.js
```diff
@@ -26,8 +26,11 @@ const mapStateToProps = state => ({
     ui: sGetUi(state),
 });
 
-const mapDispatchToProps = dispatch => ({
-    onUpdate: ui => dispatch(fromActions.fromCurrent.acSetCurrentFromUi(ui)),
+const mapDispatchToProps = (dispatch, ownProps) => ({
+    getOnClick: ui => () => {
+        dispatch(fromActions.fromCurrent.acSetCurrentFromUi(ui));
+        ownProps.onClick();
```
edoardo: This prop is not required. Like in the case of the MenuBar, is not going to be passed. Probably we need a check before calling the function.

PR #31, opened 2018-09-27
Feature/data dimension component

joakimia:
This PR provides the data visualizer with the data dimension component. This component is shown in a dialog and is used for selecting data dimensions to include in the analytic object being edited/viewed.

Refactoring and adding tests is ongoing, but I am requesting a merge so that it can have some "soak" time with the team and make the app more functional.


packages/app/src/actions/ui.js
```diff
@@ -1,4 +1,6 @@
 import { actionTypes } from '../reducers';
+//import { acSetRecommendedIds } from './recommendedIds';
```
jenniferarnesen: You can remove these commented lines

PR #32, opened 2018-09-28
Layout flex

janhenrikoverland:
- Allow for different layouts
- Implement default layout: scale axes horizontally with window size, scale axes vertically when adding dimensions and wrap to new lines, no overflow
- Extract reusable components and style to Layout folder root
- Put default layout components and style in DefaultLayout folder


packages/app/src/components/Layout/Chip.js
```diff
@@ -22,22 +26,36 @@ const styles = {
...
+            {chipLabel}
         </div>
-    ) : (
-        ''
     );
+};
+
+const Chip = ({ dimensions, itemsByDimension, axisName, dimensionId }) =>
+    dimensionId
+        ? renderChip(dimensions, itemsByDimension, axisName, dimensionId)
```
jenniferarnesen: How about just passing displayName, instead of the whole dimensions object?

packages/app/src/components/Layout/Layout.js
```diff
@@ -1,78 +1,15 @@
...
-    }
-}
-
-Layout.displayName = 'Layout';
-
-const mapStateToProps = state => ({
-    layout: sGetUiLayout(state),
-    dimensions: sGetDimensions(state),
-});
+const Layout = ({ type = 'default' }) => getLayoutByType(type);
```
jenniferarnesen: getLayoutByType takes `props` as the second argument, but a value is not provided.

PR #35, opened 2018-10-04
Component/chip

janhenrikoverland:
- Adds click handler to dimension chips
- Adds an action button/menu on the chip (stopping the click propagation so that chip click handler is not fired)
- Tab will take you from the button to the menu through aria-own

### DIFF
--- a/packages/app/src/components/Layout/Menu.js
+++ b/packages/app/src/components/Layout/Menu.js
@@ -1,0 +1,57 @@
+import React from 'react';
+import Menu from '@material-ui/core/Menu';
+import IconButton from '@material-ui/core/IconButton';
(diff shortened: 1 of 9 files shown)


packages/app/src/components/Layout/Menu.js
```diff
@@ -0,0 +1,55 @@
...
+    handleClose = event => {
+        event.stopPropagation();
+        this.setState({ anchorEl: null });
+    };
+
+    getMenuId = () => `menu-for-${this.props.id}`;
+
+    render() {
+        return (
+            <div>
```
edoardo: Fragment ?

janhenrikoverland: Yep

PR #38, opened 2018-10-08
Store metadata

janhenrikoverland:
- Adds new store prop: "metadata" (reducer, action creator, test)
- Adds middleware that listens to "metadata" props in actions


janhenrikoverland:
There is an action ADD_METADATA for that as it's not already dispatched anywhere yet.

PR #41, opened 2018-10-09
Dimension item popup

janhenrikoverland:
Shows item names in a tooltip on hover. It's a Popper rather than a Tooltip as the latter is just an html "title" attribute which does not support html or line breaks.

- show delay: 500
- hide on mouse leave, click and drag
- show "None selected" for empty dimensions

### DIFF
--- a/packages/app/src/components/Layout/Tooltip.js
+++ b/packages/app/src/components/Layout/Tooltip.js
@@ -1,0 +1,70 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Popper from '@material-ui/core/Popper';
+import Paper from '@material-ui/core/Paper';
+
+import { colors } from '../../colors';
+import { sGetUiItems } from '../../reducers/ui';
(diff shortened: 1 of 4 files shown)


packages/app/src/components/Layout/Tooltip.js
```diff
@@ -0,0 +1,69 @@
...
+        const displayNames = itemIds.length
+            ? itemIds.map(id => metadata[id] || id)
+            : [labels.noneSelected];
+
+        return displayNames.length ? this.renderTooltip(displayNames) : '';
+    }
+}
+
+const mapStateToProps = (state, ownProps) => ({
+    itemIds: sGetUiItems(state)[ownProps.dimensionId],
```
edoardo: Could it happen that this `dimensionId` is not found in that data structure?

janhenrikoverland: Pushed a default array to the render function. Will cause less re-renders than returning a new array in MSTP.

PR #48, opened 2018-10-16
Fix/util sort array

joakimia:
Change sortArray to sortObj,
Add additional parameter propName/compareName to sort on values from passed in prop name.


packages/app/src/util.js
```diff
@@ -15,10 +15,12 @@ export function entriesToObject(entries) {
     );
 }
 
-export const sortArray = array => {
-    const sortedDimensions = array.sort((dimensionA, dimensionB) => {
-        const nameA = dimensionA.displayName.toLowerCase(),
-            nameB = dimensionB.displayName.toLowerCase();
+export const sortObj = (obj, propName) => {
```
jenniferarnesen: There is no need to change this function from sortArray to sortObj. The first argument is still an array, so line 19 is redundant.  Just keep the original function, only replacing with [propName], as you did here.

PR #57, opened 2018-10-19
Task/Respect the user setting for display name property, and standardize on use of 'name' as property name

jenniferarnesen:
Changes: 
- converts the userSetting for displayProperty to the value that will be used in the apiRequest (displayName, displayShortName)
- await the fetching of userSettings and putting in the store before fetching dimensions since dimensions will use the userSettings
- rename the 'name' prop (displayName or displayShortName) to 'name' so it can consistently be used in the app
- added tests for api/dimensions


packages/app/src/api/dimensions.js
```diff
@@ -5,57 +5,68 @@ import { onError } from './index';
...
+
+// Get dimensions on startup
+export const apiFetchDimensions = nameProp => {
+    const params = `fields=id,${nameProp}~rename(name),dimensionType`;
+
+    return request('dimensions', params);
 };
 
 export const apiFetchGroups = dataType => {
+    const fields = 'fields=id,displayName~rename(name)';
```
janhenrikoverland: Data element groups do have short name (unfortunately).

packages/app/src/reducers/settings.js
```diff
@@ -27,3 +28,5 @@ export default (state = DEFAULT_SETTINGS, action) => {
 // Selectors
 
 export const sGetSettings = state => state.settings;
+export const sGetDisplayNameProperty = state =>
+    state.settings.displayNameProperty;
```
janhenrikoverland: Use the root selector :-)

PR #58, opened 2018-10-22
Use DEFAULT as default value for aggregationType

edoardo:
d2.analytics makes sure the aggregationType option is not added to the
request if the value is DEFAULT.
We need to handle the DEFAULT value for the option in the AO for backwards
compatibility.


edoardo:
This requires a fix in d2.analytics. Waiting for the new version of d2, before merging.

PR #61, opened 2018-10-22
feat(headerbar)

adeelshahid:
**Progress**

- [x] integrate at code level
- [x] styles missing from default imports
- [x] verify / test HeaderBar API integration. [PR #109](https://github.com/dhis2/ui/pull/109) should fix it for dev. environments.
- [x] translate HeaderBar app title


packages/app/src/components/App.js
```diff
@@ -85,32 +87,34 @@ export class App extends Component {
 
         return (
             <Fragment>
-                <div className="app">
-                    <div className="item1 headerbar">Headerbar</div>
-                    <div className="item2 visualization-type-selector">
-                        <VisualizationTypeSelector />
+                <UI>
+                    <HeaderBar appName="Data Visualizer" />
```
varl: Very slick.

packages/app/public/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" dir="ltr">
```
varl: Should we be able to set this from e.g. `<UI>` or do we have to do this for each app?

packages/app/src/components/App.css
```diff
@@ -33,16 +33,15 @@ body {
...
     grid-gap: 1px;
     grid-template-columns: 250px auto 385px;
-    grid-template-rows: 48px auto auto auto;
     grid-template-areas:
         'headerbar headerbar headerbar'
         'charttype menubar menubar'
         'dimensions chartlayout interpretations'
         'dimensions canvas interpretations';
+    background-color: #aaa;
+    font-family: Roboto, sans-serif;
```
varl: Isn't Roboto as the default font inherited from `<UI>`?

packages/app/public/index.html
adeelshahid: Setting this in `public/index.html` would help us let go of any `React` dependency where we keep a reference to the `<UI /> Component` itself.

It also gives us generic way to update `dir={ltr,rtl}` in all the apps.

varl: Ok.

packages/app/src/components/App.css
adeelshahid: updated, we didn't set any font on `[dir=ltr,rtl] body`. This has been fixed in `d2/ui` hence removing it.

package.json
```diff
@@ -31,5 +31,8 @@
     },
     "workspaces": [
         "packages/*"
-    ]
+    ],
+    "dependencies": {
+        "ui": "github:dhis2/ui-builds"
```
edoardo: Shouldn't this be in the app's package.json?

packages/app/src/components/App.js
```diff
@@ -96,32 +98,34 @@ export class App extends Component {
 
         return (
             <Fragment>
```
edoardo: Do we need `<Fragment>` here now that `<UI>` wraps everything?

package.json
varl: Yes.

PR #63, opened 2018-10-22
DHIS2-4971: Download menu component

edoardo:
This actually requires d2 changes (https://github.com/dhis2/d2/compare/task/filter-as-dimension-option-in-request) to make dataValueSet download links work.

Feedback appreciated on the downloadImage implementation (does not work in IE).


edoardo:
Changed approach to avoid to use the download attribute in the anchor (not supported in IE).
Instead of triggering a download directly, the image/PDF is opened in a new tab and can be seen before deciding to save it locally via the browser's Save dialog, where it's also possible to decide where to save the file.

PR #64, opened 2018-10-22
Task/org unit dimension

neeilya:
Fixes DHIS2-4450.

This PR includes changes related to integration of org-unit-dialog component.


jenniferarnesen:
@neeilya @janhenrikoverland 
Our app structure includes an api directory that contains all the api requests for the rest of the app. I think it makes sense to also move the org unit requests (and the transform function) to a file (orgUnit.js ?) in the api folder. What do you think? It would reduce the scope of responsibility of the OrgUnitDimension component.

PR #65, opened 2018-10-23
Put back onClick handler for options dialog, reordered buttons

edoardo:


edoardo:
Waiting for the changes in the store for the open/close state of the dialog before merging into master.

PR #70, opened 2018-10-25
Upgrade d2-ui deps to ones depending on d2@31

varl:


varl:
Thanks, I didn't notice that change.

PR #83, opened 2018-10-29
feature/display loading indicator and potential loading error when generating chart

jenniferarnesen:
[DHIS2-5013, DHIS2-4850]

Includes
* show loading indicator while chart is being generated
* catch errors that happen while generating chart, and display an error message
* use the full canvas to display the visualization
* loadError property in store renamed to loader, and is now an object containing the loading Error, and the loading state


packages/app/src/components/Visualization/BlankCanvas.js
```diff
@@ -1,26 +1,39 @@
...
-                    <span>{message}</span>
-                </div>
+    let canvasContent = <p style={styles.text}>{message}</p>;
+    if (loading) {
+        canvasContent = <LoadingMask />;
+    } else if (error) {
+        canvasContent = (
+            <div>
+                <img src={chartErrorImg} alt={i18n.t('Chart error')} />
+                <p style={styles.text}>{message}</p>
```
amcgee: *minor* `{message}` should just be `{error}` so we can remove the ternary from line 13 above.

PR #85, opened 2018-10-30
Fix styles for chips in layout

edoardo:
Not sure why it was broken after merging master...


jenniferarnesen:
This is why it changed: The headerbar commit introduced the UI component which wraps our app. The ui component contains a reset.css, which has box-model: border-box. Browser default is content-box.

border-box is better (imho), so its good that we go with that. Too bad the headerbar commit broke our ui though.

PR #96, opened 2018-11-02
Fix/default root org unit

neeilya:
Fix app crash on selecting existing chart.


jenniferarnesen:
@neeilya 
Could you add a step-by-step to reproduce the bug (crash)?

PR #100, opened 2018-11-05
Add InterpretationsButton component

edoardo:
I found a way to do the show/hide of the interpretations panel while expanding the grid, see the change in App.css.
It works well when no AO is loaded, but it extends the panel on the right end side of the screen when there is a chart.

I suspect the problem is due to some `width` in one of the elements inside the chart div...


edoardo:
Merged in merged-interpretations branch.

PR #101, opened 2018-11-06
DHIS2-3208: Interpretations panel and button

edoardo:


packages/app/src/components/Interpretations/Interpretations.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import InterpretationsComponent from '@dhis2/d2-ui-interpretations';
+
+export class Interpretations extends Component {
+    constructor(props, context) {
+        super(props);
+
+        this.d2 = context.d2;
```
jenniferarnesen: Is it necessary to set a class property or can it just use this.context.d2 in the render fn?

edoardo: I've refactored the component into a functional one.

PR #103, opened 2018-11-06
CSS grid fixes

edoardo:
To make the CSS grid collapse properly when the interpretations panel is toggled, the container of highchart must be absolute positioned.
This caused a minor overflow problem where the legend of the chart was outside the parents' container boundary. That was due to the container having also the TitleBar component.
To solve the problem I added a new grid cell for the TitleBar.
Then the grid-gap was showing a line below the TitleBar also when there was no title.
By replacing grid-gap with the required border rules for the various cells, the undesired lines are removed.

### DIFF
--- a/packages/app/src/components/App.js
+++ b/packages/app/src/components/App.js
@@ -123,8 +123,10 @@ export class App extends Component {
                     <div className="item6 interpretations">
                         Interpretations panel
                     </div>
-                    <div className="item8 canvas">
+                    <div className="title-bar">
                         <TitleBar />
+                    </div>
+                    <div className="item8 canvas">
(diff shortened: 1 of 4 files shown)


packages/app/src/components/App.js
```diff
@@ -123,8 +123,10 @@ export class App extends Component {
                     <div className="item6 interpretations">
                         Interpretations panel
                     </div>
-                    <div className="item8 canvas">
+                    <div className="title-bar">
                         <TitleBar />
+                    </div>
+                    <div className="item8 canvas">
```
jenniferarnesen: you probably don't want `item8`, right?

edoardo: no, but those are already removed in the other branch

PR #108, opened 2018-11-09
Task/dimension dialogs addtolayout button [DHIS2-5162]

joakimia:
This PR enables opening of unselected dimension dialogs.
When opening an unselected dimension, the "Update" button is gone,
instead a dropdown button with "Add to series", "Add to category" "Add to filter" is shown.

If the layoutType is equal to yearOnYear, then only "Add to filter" will be rendered.

The Button loops through ui.layout to check if the dialog ID is present.
If the ID is present, the Dimension is selected and the regular "Update" button will render.


packages/app/src/components/Dimensions/DimensionOptions.js
```diff
@@ -56,20 +43,18 @@ export class DimensionOptions extends Component {
     };
 
     getMenuItems = () =>
-        items.map(option => (
+        ADD_TO_LAYOUT_OPTIONS.map(option => (
             <MenuItem
                 key={option.axisName}
+                children={i18n.t(option.name)}
```
amcgee: Same as above, don't translate a variable.

packages/app/src/modules/layout.js
```diff
@@ -15,6 +15,13 @@ export const AXIS_NAMES = [
 export const DIMENSION_ID_PROP_NAME = 'dimension';
 export const DIMENSION_ITEMS_PROP_NAME = 'items';
 
+// Keys and displayName for adding dimensions to layout
+export const ADD_TO_LAYOUT_OPTIONS = [
+    { axisName: 'columns', name: 'Add to series' },
```
amcgee: Do the translating here

packages/app/i18n/en.pot
```diff
@@ -59,15 +59,6 @@ msgstr ""
...
 msgstr ""
 
-msgid "Add to series"
-msgstr ""
-
-msgid "Add to category"
-msgstr ""
-
-msgid "Add to filter"
-msgstr ""
```
amcgee: These should be re-generated

PR #109, opened 2018-11-09
Latest data viz design

janhenrikoverland:


packages/app/i18n/en.pot
```diff
@@ -5,8 +5,13 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
+<<<<<<< HEAD
+"POT-Creation-Date: 2018-11-08T18:44:19.452Z\n"
+"PO-Revision-Date: 2018-11-08T18:44:19.452Z\n"
+=======
 "POT-Creation-Date: 2018-11-07T21:50:49.328Z\n"
 "PO-Revision-Date: 2018-11-07T21:50:49.328Z\n"
+>>>>>>> e8b0da797b90aefca1acc02c5086025dcca15dc0
```
amcgee: Merge conflict should be removed

packages/app/src/assets/DataIcon.js
```diff
@@ -1,74 +1,119 @@
...
+                            />
+                        </g>
                     </g>
                 </g>
             </g>
         </svg>
     );
 };
 
+// const DataIcon = ({ style }) => {
```
amcgee: Probably should just remove commented out lines, but that can also be done later

packages/app/src/components/Dimensions/DimensionItem.js
```diff
@@ -57,19 +69,21 @@ export class DimensionItem extends Component {
                 onMouseOver={this.onMouseOver}
                 onMouseLeave={this.onMouseExit}
             >
-                <DimensionLabel {...this.props}>
-                    {Icon}
+                <DimensionLabel {...this.props} isDeactived={this.isDeactived}>
+                    <div style={styles.iconWrapper}>{Icon}</div>
```
amcgee: Ideally these should be `className` instead of `style` (better for performance and flexibility) but we need to update that a lot more places.  It would be good to avoid adding new `style` attribute usage.

PR #113, opened 2018-11-09
Task/Use d2-ui theme

amcgee:
Use the theme from d2-ui and override menuItem padding, fontSize, and transition.

This will automatically make dialogs look a lot nicer when combined with the latest `d2-ui` v5.0.2

The MenuItem overrides could probably be moved to d2-ui, but leaving them here for now to minimize change surface are.


amcgee:
Note:

This reduces font size and padding of Mui MenuItems globally using the theme.  However, it does not change the font size in MenuItems which contain a ListItemText (such as the FileMenu).  In those cases, ListItemText wraps the text in a `<Typography variant="subtitle" />` element which resets the fontSize to 100% (16px).  We could try to change this behavior in the `d2-ui` FileMenu and other locations, but for now I think it's enough to reduce the padding and restrict this change to `data-visualizer`

The next step would be to take the theme customizations from this change and include them in a new publish of `d2-ui`.  I could alternatively do that step first, looking for feedback on which order to do them in @jenniferarnesen @edoardo @janhenrikoverland

amcgee:
Question - why is diff-coverage failing?  Codeclimate doesn't seem to show a reason

PR #116, opened 2018-11-12
Fixes for running app once installed from app store

edoardo:
The `scriptPrefix` in webpack config is probably not going to work for the app bundled in core.


packages/app/config/webpack.config.prod.js
```diff
@@ -59,7 +59,7 @@ const extractTextPluginOptions = shouldUseRelativeAssetPaths
       { publicPath: Array(cssFilename.split('/').length).join('../') }
     : {};
 
-const scriptPrefix = '..';
+const scriptPrefix = '../../..';
```
jenniferarnesen: Instead of hardcoding this, it can probably be picked up from the manifest activities.dhis.href property

jenniferarnesen:
Regarding the yarn.lock file, doesn't seem like the har-validator info got updated with the correct gzipped tarball.  Try this: ```yarn cache list --pattern har-validator```

PR #120, opened 2018-11-14
Fix/dimension dialog searchfield [DHIS2-5212]  

joakimia:
This PR includes:
-fixing filter function in the Data Dimension and Dynamic Dimension dialogs to filter on items correctly (did not filter at all before).
- Adding override rules to theme.js to change the underline style to mirror the dashboard application.
- Adding  folder "Filter", similar to the dashboard app and reusing this in Dimensions.js / DataDimension.js / DynamicDimension.js instead of importing TextFields individually.
- Adding an IconButton as endAdornment to the TextField to clear the input field onClick.
- Adding Escape key support for resetting the input field, both in Dimensions.js and the dialogs.


packages/app/src/components/Filter/Filter.js
```diff
@@ -0,0 +1,73 @@
...
+const onChangeWrapper = (props, event) => {
+    event.target.value.length
+        ? props.onChange(event.target.value)
+        : props.onClear();
+};
+
+const onKeyDownWrapper = (onClear, event, text) => {
+    if (event.key === 'Escape') {
+        event.preventDefault();
+        !text.length ? ref && ref.blur() : onClear();
```
jenniferarnesen: Does it work incorrectly if you write it like this?

```!text.length && ref ? ref.blur() : onClear();```

jenniferarnesen: The difference being that onClear() will be called if ref is null.

PR #124, opened 2018-11-15
Fix/dimension context menu [DHIS2-5211]

joakimia:
This PR includes fixes to the context menu for dimensions in  left pane:

**if current layout is Year on Year, menu shows:**
- unselected: "Add to filter"
- selected: "Remove"

**if current layout is not Year on Year, context menu shows:**
- unselected: "Add to series/category/filter"
- selected: "Move to [layout not currently selected]" + "Remove"

I've moved the Context menu component outside the dimension label (instead of being rendered as a child) to not cause any propagation issues.

**Theres additional refactoring in this PR (apologies):**
- Removing unused styling after the new UI change
- Removing unused props.
- Removing unused code (e.g RemoveDimensionButton)
- Adding more test cases and some changes to existing ones.

**Additional Notes:**
- The context menu dismounts a bit slow (because of animation)? resulting in the menu to render  "Move To" during onClose/dismount.

Sometimes, If i spam the MenuItem, the alternatives have changed from, "add to" to "move to".
If i double click fast enough, the dimension can potentially be added to the given layout, and then the second click will move it to the alternative that is being re-rendered (try it yourself).

Im open to feedback/pointers to fix this (might happen often on wooden computers).
A work-around could be to assign the onDoubleClick as well (?), or override some transition styling on the Menu component? 

Ive tried using the other callbacks from the Material-UI's Menu component (onExit, onExiting etc.).
And tried to have a dedicated boolean instead of the anchorEl for the "open" prop. But the same behaviour persists.

### DIFF
--- a/packages/app/src/components/Dimensions/DimensionLabel.js
+++ b/packages/app/src/components/Dimensions/DimensionLabel.js
@@ -1,36 +1,24 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-
-import Close from '@material-ui/icons/Close';
-import {
-    acRemoveUiLayoutDimensions,
-    acSetUiActiveModalDialog,
-    acSetUiItems,
-} from '../../actions/ui';
-import { sGetUiLayout, sGetUiItems } from '../../reducers/ui';
-
+import { acSetUiActiveModalDialog } from '../../actions/ui';
 import { styles } from './styles/DimensionLabel.style';
-
-export const RemoveDimensionButton = ({ action }) => (
-    <button style={styles.deleteButton} onClick={action} tabIndex={0}>
-        <Close style={styles.deleteButtonIcon} />
-    </button>
-);
 
 export class DimensionLabel extends Component {
     static propTypes = {
-        openDialog: PropTypes.func.isRequired,
         id: PropTypes.string.isRequired,
+        type: PropTypes.string.isRequired,
         name: PropTypes.string.isRequired,
-        isSelected: PropTypes.bool.isRequired,
         isDeactivated: PropTypes.bool.isRequired,
-        removeDimension: PropTypes.func.isRequired,
+        isSelected: PropTypes.bool.isRequired,
+        openDialog: PropTypes.func.isRequired,
         children: PropTypes.arrayOf(PropTypes.element).isRequired,
     };
 
     onLabelClick = () => {
-        this.props.openDialog(this.props.id);
+        if (!this.props.isDeactivated) {
+            this.props.openDialog(this.props.id);
+        }
     };
 
     onKeyPress = event => {
@@ -45,7 +33,7 @@ export class DimensionLabel extends Component {
             onClick={this.onLabelClick}
             onKeyPress={this.onKeyPress}
             tabIndex={0}
-            style={styles.unselected}
+            style={styles.label}
         >
             {this.props.children}
         </div>
@@ -54,29 +42,13 @@ export class DimensionLabel extends Component {
     render = () => {
         const Label = this.renderLabel();
 
-        const containerStyle =
-            this.props.isSelected && !this.props.isDeactivated
-                ? styles.selected
-                : styles.unselected;
-
-        return (
-            <div className="labelContainer" style={containerStyle}>
-                {Label}
-            </div>
-        );
+        return Label;
     };
 }
 
-const mapStateToProps = state => ({
-    currentLayout: sGetUiLayout(state),
-    items: sGetUiItems(state),
-});
-
 export default connect(
-    mapStateToProps,
+    null,
     {
         openDialog: id => acSetUiActiveModalDialog(id),
-        removeDimension: id => acRemoveUiLayoutDimensions(id),
-        setUiItems: items => acSetUiItems(items),
     }
 )(DimensionLabel);
--- a/packages/app/src/components/Dimensions/DimensionOptions.js
+++ b/packages/app/src/components/Dimensions/DimensionOptions.js
@@ -1,61 +1,111 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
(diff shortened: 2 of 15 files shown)


packages/app/src/components/Dimensions/DimensionOptions.js
```diff
@@ -1,61 +1,111 @@
...
+                  FILTER,
+                  this.addDimension,
+                  `Add to ${axisLabels[FILTER]}`
+              )
+            : Object.entries(axisLabels).map(([key, label]) =>
+                  this.renderMenuItem(
+                      `add-to-${key}`,
+                      key,
+                      this.addDimension,
+                      `${i18n.t('Add to')} ${label}`
```
edoardo: For this type of strings, is best to use variables within the string, as different languages might need to put the label part in a different place in the string.
Use this format instead: `${i18n.t('Add to {label}', { label })`.

packages/app/src/components/Dimensions/DimensionOptions.js
```diff
@@ -1,61 +1,111 @@
...
+        this.props.removeDimension(id);
     };
 
+    getAddToItems = () => [
+        isYearOverYear(this.props.type)
+            ? this.renderMenuItem(
+                  `add-to-${this.props.id}`,
+                  FILTER,
+                  this.addDimension,
+                  `Add to ${axisLabels[FILTER]}`
```
edoardo: Missig i18n, see below.

packages/app/src/components/Dimensions/DimensionOptions.js
```diff
@@ -1,61 +1,111 @@
...
+        const layout = Object.entries(this.props.currentLayout);
+
+        return layout
+            .filter(([axisKey, axisIds]) => !axisIds.includes(this.props.id))
+            .map(([axisKey, axisIds]) =>
+                this.renderMenuItem(
+                    `${this.props.id}-to-${axisKey}`,
+                    axisKey,
+                    this.addDimension,
+                    `${i18n.t('Move to')} ${axisLabels[axisKey]}`
```
edoardo: See comment above

packages/app/src/components/Dimensions/DimensionLabel.js
```diff
@@ -54,29 +42,13 @@ export class DimensionLabel extends Component {
     render = () => {
```
jenniferarnesen: Just rename the renderLabel fn to render, and get rid of this render fn. No need to break it up like it is currently.

packages/app/src/components/Dimensions/DimensionLabel.js
```diff
@@ -54,29 +42,13 @@ export class DimensionLabel extends Component {
...
-const mapStateToProps = state => ({
-    currentLayout: sGetUiLayout(state),
-    items: sGetUiItems(state),
-});
-
 export default connect(
-    mapStateToProps,
+    null,
     {
         openDialog: id => acSetUiActiveModalDialog(id),
```
jenniferarnesen: Here you can use the shorthand notation:
`{openDialog: acSetUiActiveModalDialog}`

packages/app/src/components/Dimensions/DimensionOptions.js
```diff
@@ -1,61 +1,124 @@
...
         this.setState({ anchorEl: null });
-        this.props.onClose();
     };
 
     addDimension = axisName => {
         this.props.onAddDimension({ [this.props.id]: axisName });
-        this.handleClose();
-        setTimeout(() => this.props.openDialog(this.props.id), 10);
+
+        const items = this.props.items[this.props.id];
```
jenniferarnesen: You can simplify this by using the recently merged code (see mapStateToProps in DynamicDimension.js)
```
items: sGetItemsByDimension(state, ownProps.id) || emptyItems
```

Then all you need to do here is
```
if (!this.props.items.length) ...
```

packages/app/src/components/Dimensions/RecommendedIcon.js
```diff
@@ -39,7 +35,7 @@ export class RecommendedIcon extends Component {
             ? this.showTooltip()
             : null;
 
-        return this.checkIfRecommended() ? (
+        return Boolean(this.props.isRecommended) ? (
```
jenniferarnesen: Is it necessary to cast to Boolean?  isRecommended comes from javascript array.includes, which returns a boolean already

packages/app/src/components/Dimensions/__tests__/DimensionItem.spec.js
```diff
@@ -56,4 +57,15 @@ describe('The DimensionItem component ', () => {
 
         expect(dimOptions.length).toEqual(1);
     });
+
+    it('renders a DimensionLabel with additional styling properties if props isSelected = true', () => {
```
jenniferarnesen: I think you mean DimensionItem, not DimensionLabel, in the test title?

packages/app/src/components/Dimensions/__tests__/DimensionOptions.spec.js
```diff
@@ -37,4 +48,19 @@ describe('The DimensionOptions component ', () => {
 
         expect(dropDown.length).toEqual(1);
     });
+
+    it('passes only 1 element ("Add to Filter", or "Remove") as menuItems to <DropDown /> if prop isSelected is true and prop type is equal to YearOnYear', () => {
```
jenniferarnesen: Remove the test inside the parentheses, since this text is not actually checked. Same goes for the next test

packages/app/src/modules/layout.js
```diff
@@ -24,6 +24,12 @@ export const ADD_TO_LAYOUT_OPTIONS = [
     { axisName: 'filters', name: i18n.t('Add to filter') },
 ];
 
+export const axisLabels = {
```
jenniferarnesen: Good idea to export this. Then you can also use it in DefaultAxis.js and YearOverYearAxis.js

PR #138, opened 2018-11-22
Fix scrolling in FF/Safari

janhenrikoverland:
- Workaround for vertical flex scrolling in Firefox/Safari
- Fixes broken scrolling for Dimensions og Interpretations


jenniferarnesen:
Looks good.  A good commit title could be:

Fix missing vertical scrollbars on dimension and interpretation side panels in FF/Safari

PR #139, opened 2018-11-22
DHIS2-5257: snackbar handling of info/warning/error messages

edoardo:
See: https://docs.google.com/document/d/1LHh29S99FBPfiKmUy7JwNeAZkIPbiG8QsezOAL2lnjA/edit


packages/app/src/components/App.js
```diff
@@ -140,16 +139,7 @@ export class App extends Component {
     }
 
     renderSnackbar() {
```
jenniferarnesen: You can get rid of the renderSnackbar function and just include `<Snackbar />` directly in render fn

edoardo: Fixed

jenniferarnesen:
Would be nice to get rid of the black snackbar that is briefly seen when closing the warning.  Btw, the warning didn't close on its own now, I had to click the X.

PR #144, opened 2018-11-26
Refactor/dimension dialogs render structure/styling rules

joakimia:
This PR includes:
changes to styling:
- use flex to dynamically render / position components.
- use classNames instead of inline-styling.
- remove unecessary width / height rules of sub-components.
- other minor stuff ( remove unecesarry arrow function on lifecycle methods ).

- Moving Select button from UnselectedItems.js to DataDimension.js / DynamicDimension.js
- Moving Deselect button from SelectedItems.js to DataDimension.js. / DynamicDimension.js

Along with the buttons, the arrays containing the highlighted ids of Unselected / Selected dimensions have been moved up to DataDimension.js / DynamicDimension.js.

The current Data Dimension renders the Select / Deselect button with position absolute and left / right (with pixel units, i have also tried percentages) relative to its parent container. Since the containers have different sizes, the Dialog flexes incorrectly, resulting in mis-aligned  Select / De-select buttons.

The DataDimension.js have increased in size but in order to render the components in such fashion that flexing behaves properly i had to structure the  UnselectedtItems, Buttons and SelectedItems as siblings.

I have only moved the necessary functionality required for the Select / Deselect buttons onClick handler up to the DataDimension ( passing the array of highlighted Ids to the button's  onClick fn).

This also correspond with the structure of the Period-Selector, which will help creating consistency with the Dimension dialogs.

If you think the DataDimension is too bulky to be expaneded upon, we can change the positioning to use percentages instead of pixel units. 
 
There's also some minor improvements (checking length of highlighted Ids before dispatching the ADD / REMOVE_ITEMS ).

Current behavior of DataDimension (pixel units)
image

Small:
image

With percentages:
image

image

(To highlight the principle/issue):
image

New Approach:

Normal:
image

Small:
image

Smaller:
image


joakimia:
Ill close this PR and keep the local branch as this is not a critical change and only includes a refactor of the Data dimension compoent.

PR #153, opened 2018-11-28
Catch fatal React errors with a full-screen error boundary

amcgee:
Same change as [in the Maps app](https://github.com/dhis2/maps-app/pull/65)

I will create tests for the error boundary in the morning.

### DIFF
--- a/packages/app/src/components/ErrorBoundaries/FatalErrorBoundary.js
+++ b/packages/app/src/components/ErrorBoundaries/FatalErrorBoundary.js
@@ -1,0 +1,179 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import i18n from '@dhis2/d2-i18n';
+import { withStyles } from '@material-ui/core/styles';
+import InfoIcon from '@material-ui/icons/InfoOutlined';
+
+const bgColor = '#F4F6F8',
+  secondaryTextColor = '#494949';
+
+const replaceNewlinesWithBreaks = text =>
+  text
+    .split('\n')
+    .reduce((out, line, i) => [...out, line, <br key={i} />], []);
+
+const styles = {
+  mask: {
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    right: 0,
+    bottom: 0,
+
+    overflow: 'auto',
+    overflowY: 'auto',
+
+    backgroundColor: bgColor,
+
+    display: 'flex',
+
+    minWidth: 640,
+    minHeight: 480,
+
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  container: {
+    textAlign: 'center',
+    color: 'black',
+  },
+  icon: {
+    width: 96,
+    height: 96,
+    color: '#B0BEC5',
+    marginBottom: 24,
+  },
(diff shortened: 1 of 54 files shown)


packages/app/src/components/ErrorBoundaries/FatalErrorBoundary.js
```diff
@@ -0,0 +1,179 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import i18n from '@dhis2/d2-i18n';
+import { withStyles } from '@material-ui/core/styles';
+import InfoIcon from '@material-ui/icons/InfoOutlined';
+
+const bgColor = '#F4F6F8',
```
jenniferarnesen: we have these colors defined in colors.js "snow" and  "greyDark". (I know, great names...)

packages/app/src/components/ErrorBoundaries/FatalErrorBoundary.js
```diff
@@ -0,0 +1,179 @@
...
+                            className={
+                                this.state.drawerOpen
+                                    ? classes.drawerVisible
+                                    : classes.drawerHidden
+                            }
+                        >
+                            <div className={classes.errorIntro}>
+                                {i18n.t(
+                                    'An error occurred in the DHIS2 {{AppName}} application.',
+                                    { AppName: 'Data Visualizer' }
```
edoardo: We do translate the app name in the headerbar, so perhaps we can remove the variable so that the whole string can be translated with the app name in it.

packages/app/src/components/ErrorBoundaries/FatalErrorBoundary.js
amcgee: Yeah, they're also in the theme but I wanted to keep dependencies to a minimum so I decided to hard-code them.  I'll import them directly from the mui3.there object (not referencing the mui3 theme) since that's becoming the canonical source of truth and the tests should catch any issues with dereferencing those variables.

packages/app/src/components/ErrorBoundaries/FatalErrorBoundary.js
```diff
@@ -0,0 +1,179 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {
+            error: null,
+            errorInfo: null,
+            drawerOpen: false,
+        };
+    }
+
+    componentDidCatch(error, info) {
```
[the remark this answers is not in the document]
amcgee: Done

packages/app/src/components/ErrorBoundaries/FatalErrorBoundary.js
amcgee: Done

amcgee:
@jenniferarnesen @edoardo updated and added simple tests, wasn't sure if there was more I should test but this should cover this basics.

amcgee:
Snapshot test isn't system-agnostic, fixing...

amcgee:
OK, tests should pass now!

amcgee:
This is ready to go, any more suggestions @jenniferarnesen?

PR #160, opened 2018-11-29
fix/add url-polyfill for URLSearchParams

jenniferarnesen:
n IE, opening a file from the file->menu crashes. This was due to URLSearchParams not being supported in IE. Solution was to add the url-polyfill. 

url-polyfill is used in maps-app to solve the same problem.


jenniferarnesen:
wrong branch

PR #166, opened 2018-11-30
fix/validate selected groups before rendering placeholder in "Event Data Items" and "Program Indicators" [DHIS2-5452]

joakimia:
This PR includes:
check if user have selected a group value in "Event Data Items" and "Program Indicators" before rendering the placeholder.

Currently, user are able to change Program, but the placeholder will be rendered regardless


neeilya:
Could you please describe steps to reproduce the bug?

joakimia:
Steps to reproduce:
**1.** Open Data Dimension
**2.** Select Data type: Event Data Items / Program Indicator
**3.** Select any program.

Current behaviour (bug): Alternatives are rendered in the available dimension item list - But Selected Program still show the placehoder "Select Program".

Correct behaviour: Dropdown value should change to the selected program clicked in step 3.

PR #168, opened 2018-12-03
DHIS2-5434: fix flashing chart when loading interpretation directly from URL

edoardo:
The PR also removes unnecessary analytics requests for the latest chart's data.


packages/app/src/reducers/visualization.js
```diff
@@ -19,3 +19,5 @@ export default (state = DEFAULT_VISUALIZATION, action) => {
 // Selectors
 
 export const sGetVisualization = state => state.visualization;
+export const sGetInterpretations = state =>
+    state.visualization ? state.visualization.interpretations : [];
```
jenniferarnesen: here you can use `sGetVisualization(state).interpretations : [];`

edoardo: Hmm... default for visualization is {null}, and I think it was changed from `{}`.
I would leave that as it is.

packages/app/src/components/Visualization/__tests__/Visualization.spec.js
```diff
@@ -247,11 +247,10 @@ describe('Visualization', () => {
         vis.instance().recreateChart = recreateChartFn;
 
         it('triggers a reflow when rightSidebarOpen prop changes', () => {
-            vis.setProps({
-                ...props,
-                current: undefined,
-                rightSidebarOpen: true,
-            });
+            // simulate prevProps ?!
```
jenniferarnesen: would be great to get a test on the new logic

edoardo: The tests have been updated to account for the new logic, in particular the ones for tDoLoadvisualization.

PR #173, opened 2018-12-04
Subscribe to analytic object id to remount org units selector

neeilya:
This PR includes bug-fix for org units dimension.

Steps to reproduce (do not refresh the page during the process):
1) Open AO with preselected org units (multiple levels)
2) Open org units dimension dialog
3) Open another AO
4) Open org units dimension dialog again
5) Previously expanded org units are not selected anymore, but they are still expanded

Correct behaviour - only selected org units should be expanded.

I had a discussion with @joeDHIS about keeping org units expanded during opening/closing the dialog.
Although it is not the exact use case, the main idea was that whenever user chooses some org units and then clicks "deselect all" or reopens the dialog, org units should stay expanded (so that user wouldn't have to expand them again) even if they are not selected.

This applies some limitations to implementation, in particular - inability to tweak `org-units-tree` d2-ui component so that it would automatically collapse unselected org units. It is technically possible and I have implemented it locally, but preserving org units opened on AO change would require passing AO id as a prop all the way down to `org-unit-tree` component.

Current implementation uses sort of "subscribe" approach and re-renders `org-units-selector` component only when user changes AO (still uses lazy mount approach, so reopening the dialog does not cause additional requests). It does not repeat network requests and only fetches data for org units not expanded before.

### DIFF
--- a/packages/app/src/components/Dimensions/Dialogs/OrgUnitDimension/OrgUnitDimension.js
+++ b/packages/app/src/components/Dimensions/Dialogs/OrgUnitDimension/OrgUnitDimension.js
@@ -76,7 +76,7 @@ export class OrgUnitDimension extends Component {
         const previousId = prevProps.current ? prevProps.current.id : null;
         const currentId = this.props.current ? this.props.current.id : null;
 
-        // remount org units selector component to esnure
+        // remount org units selector component to ensure
         // only selected org units are expanded
         if (previousId !== currentId) {
             this.hideOrgUnitsTree();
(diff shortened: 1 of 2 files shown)


packages/app/src/components/Dimensions/Dialogs/OrgUnitDimension/OrgUnitDimension.js
```diff
@@ -70,6 +72,27 @@ export class OrgUnitDimension extends Component {
         this.loadOrgUnitGroups();
     }
 
+    componentDidUpdate(prevProps) {
+        const previousId = prevProps.current ? prevProps.current.id : null;
+        const currentId = this.props.current ? this.props.current.id : null;
+
+        // remount org units selector component to esnure
```
janhenrikoverland: "ensure" typo

PR #176, opened 2018-12-05
Fix/ie11 numberfield [DHIS2-5281]

joakimia:
THIS PR Includes adding new file NumberBaseOption.js and enabling increment / decrement on number fields in Internet Explorer:

The TextField is similar to TextBaseOption but have a new prop "onKeyDown" to catch the arrowUp or arrowDown input events.

On arrow Up or Left i am preventing the default behaviour ( Textfield firing the onChange callBack ) and incrementing / decrementing manually.
 
- Since the value is an empty string, i had to add a terniary check to increment/decrement the empty string with 0. 

- I also had to parse the value to int, then back to string.

- If key event is not an ArrowLeft or Right, the event propagates to the onChange callback which includes checking the type (text or number) and handling the input accordingly.

Additional info:
There are some limitations with both the TextField in general, and input components of type='number' in Internet Explorer. Currently i've only found nasty hacks but no concrete and neat solutions.

The user are still able to type in characters in internet explorer (I am not sure why, as the input has type='number' and i am not able to find solutions or relevant information on google regarding this behaviour) - but I think this is because the input starts as an uncontrolled component due to the value being an empty string (?)

I have tried to add additional checks to ignore characters but the input field still receives A-F as valid (however, only on the first input, after adding the onChange wrapper):

```
const onChangeWrapper = (value, props) => {
   // check if type is number
    if (isNumberField(props.type)) {

      // check if input is a valid Number, this does not work on the first input in IE11
        !isNaN(value) && props.onChange(value);

    } else {
      // type == 'text' , pass the change
        props.onChange(value);
    }
};

```

The Material-UI folks simply states that the TextField component is already too complex and it can't handle all the limitations from IE.

I have tried to use InputProps / inputProps but the issue remains.
 
Also, when using the mathematical number 'e' and then increments, it resets the field (both in chrome and in IE11) - but i assume this is rarely if not never used.

I looked into using [other Material-UI components](https://material-ui.com/demos/text-fields/#formatted-inputs) to format the input but this involved adding third library packages like `react-text-mask` and `react-number-format`.

But since the idea of this PR relates to increment / decrement. i chose to open a pull request with only these changes.

However, do not hesitate to request changes!

### DIFF
--- a/packages/app/src/components/VisualizationOptions/Options/BaseLineLabel.js
+++ b/packages/app/src/components/VisualizationOptions/Options/BaseLineLabel.js
@@ -4,7 +4,6 @@ import i18n from '@dhis2/d2-i18n';
 
 const BaseLineLabel = () => (
     <TextBaseOption
-        type="text"
         option={{
             name: 'baseLineLabel',
             label: i18n.t('Base line title'),
--- a/packages/app/src/components/VisualizationOptions/Options/BaseLineValue.js
+++ b/packages/app/src/components/VisualizationOptions/Options/BaseLineValue.js
@@ -1,10 +1,9 @@
 import React from 'react';
-import TextBaseOption from './TextBaseOption';
+import NumberBaseOption from './NumberBaseOption';
 import i18n from '@dhis2/d2-i18n';
 
 const BaseLineValue = () => (
-    <TextBaseOption
-        type="number"
+    <NumberBaseOption
         option={{
             name: 'baseLineValue',
             label: i18n.t('Base line value'),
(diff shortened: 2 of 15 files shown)


joakimia:
I am keeping this change at a local branch and will possibly take new shot at this on the Data vis v1 epic.

After discussion with Edoardo and Jan on Slack, we agreed to close this PR.

The Idea was to add possibility of incrementing / decrementing the number field, with arrow keys (as the Up/Down buttons does not render in IE - instead an X icon will be rendered, indicating that the type='number' is not evaluated correctly).  This was OK but more limitations on the textField in IE11 was discovered.

Given that the value is initially an empty string, the input field in IE11 starts as an uncontrolled component. No conditional checks is catching the first character - meaning that the user can type in invalid values.

Additionally, the pattern/mask attribute  did not filter invalid values based on the RegEx input.

After testing in IE11 we found several limitations which meant that we had to add more browser specific code while simultaneously not change the behaviour in chrome / firefox. This resulted in a quite complex TextField component.

I found no concrete solutions that could fit in into our current codebase (im sure it exists but we agreed not to spend more time on it) - because of that we chose to leave this one out as the user can still type in numbers and perform the needed task without the increment/decrement functionality added in this PR.

PR #188, opened 2018-12-07
Layout validation

janhenrikoverland:
- Cancels "update" before requests/render if store.current does not meet certain conditions.
- Makes it impossible to unselect all YOY series items.


edoardo:
If I unselect everything in the orgunit selector, the analytics response is valid, so the chart is shown, but it's not clear to which orgunit the data refers to.

PR #204, opened 2019-01-22
Data items

janhenrikoverland:
- Fetches programDataElements and programTrackedEntityAttributes for "Event data items"
- Fetches programIndicators for "Program indicators"
- Adds support for a selector function that can be used to get the right stuff from the response when it's not straight forward (not just response[entitity])
- Joins, filters and sorts PDE and PTEA

### DIFF
--- a/packages/app/src/modules/array.js
+++ b/packages/app/src/modules/array.js
@@ -1,4 +1,1 @@
 export const toArray = param => (Array.isArray(param) ? param : [param]);
-
-export const createSortFunction = key => (a, b) =>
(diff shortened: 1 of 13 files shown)


packages/app/src/modules/array.js
```diff
@@ -1 +1,4 @@
 export const toArray = param => (Array.isArray(param) ? param : [param]);
+
+export const createSortFunction = key => (a, b) =>
```
jenniferarnesen: I'm pretty sure that lodash sortBy or orderBy will do this for you.

PR #205, opened 2019-01-29
Fetch chart always by id when used in dashboard

edoardo:
This is to avoid flashing of the chart in dashboard.
The change allows for the config object to have several keys, but still
fetch the chart by id as the config object passed from dashboard to the
plugin is a stripped down version of the AO and does not have the
necessary data for generating the chart without fetching it.

### DIFF
--- a/packages/plugin/src/ChartPlugin.js
+++ b/packages/plugin/src/ChartPlugin.js
@@ -95,7 +95,7 @@ class ChartPlugin extends Component {
 
         try {
             const visualization =
-                Object.keys(config).length === 1
+                forDashboard || Object.keys(config).length === 1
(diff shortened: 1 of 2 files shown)


packages/plugin/src/ChartPlugin.js
```diff
@@ -95,7 +95,7 @@ class ChartPlugin extends Component {
 
         try {
             const visualization =
-                Object.keys(config).length === 1
+                forDashboard || Object.keys(config).length === 1
```
jenniferarnesen: could you add a unit test for when forDashboard is true?

edoardo: Added.

PR #206, opened 2019-01-31
[WIP] Feature/multi drag dimensions

jenniferarnesen:
Feature: enable user to sort the selected items list by dragging multiple items at the same time. This feature should be available in the Dynamic Dimensions, Data Dimension, and Period Dimension dialogs. In order to share code/remove duplication, since Period Dimension is a d2-ui component, it was necessary to separate out the common functionality into a new shared component "Item Selector". This shared component is currently home in the data-visualizer-app's new "shared" package. This may change as we evolve the structure of the repositories.

Changes included in this PR:
- Create separate package `shared` that can contain common modules and components
- Factor out the ItemSelector component and put it in `shared`
-- note that the ItemSelector component will be used by d2-ui PeriodSelector
- Implement multi-drag functionality in `ItemSelector` based on the react-beautiful-dnd package
- Move the `colors` module to shared
- Change DialogManager so that only the orgunit dimension component is retained in the DOM after the dialog is closed. This was to fix an issue where the drag/drop left the data and dynamic dimension selected items in an invalid state when dropping outside of the dialog. (the reason for keeping orgunit in the DOM is so that all the org unit data doesn't have to be refetched - lots of queries. There is a separate discussion on how to do this better than showing/hiding components)

Fixes [DHIS2-5480]


janhenrikoverland:
This feature has been smoke tested and we will merge it for the sake of getting started on the new mono repo. Needs more testing before being released or added to d2-ui.

PR #214, opened 2019-02-11
fix: use lerna to share code instead of copying resources

varl:
Using Lerna it is not necessary to duplicate the source code from `data-visualizer-app` into `data-visualizer-plugin` to share code between the modules.

One can just refer to the package name in the import statement to re-use code.

This works even if a package is not published to NPM since Lerna creates symlinks between all the packages in `packages/`, and using Yarn Workspaces, these links are hoisted to the top-level `node_modules/`:

```
.../apps/data-visualizer-app/node_modules > ls -l | grep data-visualizer
data-visualizer-app -> ../packages/app/
data-visualizer-plugin -> ../packages/plugin/
```

The one gotcha in `packages/plugin` is that one needs to run `yarn build` in that package before the App can find the paths. This could be automated to be run on file changes, and it would be possible to use Lerna to run `yarn start` in both packages at the same time.


jenniferarnesen:
In our discussion yesterday we talked about extracting a 3rd package in datavis called 'shared'. To me, the relevant code in this PR would go in that and both the app and plugin would import from there:

`import { apiFetchVisualization } from 'data-visualizer-shared/src/api/visualization';`

edoardo:
LGTM.
We can move the shared parts into it's own package as @jenniferarnesen suggested.

varl:
@jenniferarnesen, @edoardo: Do you want me to do it in this PR or would you want to do it yourselves later?

jenniferarnesen:
@varl we can take it over, thanks for the help!

PR #232, opened 2019-04-02
fix: Fetching analytics for analytical object with undefined aggregationType in plugin

neeilya:
### Problem:
When we pass full analytical to the plugin without `aggregationType` property specified, it passes it to analytics request.

As a result we have request like this:
_localhost:8080/api/29/analytics.json?dimension=dx:ntzmpYRSKGg&dimension=ou:ImspTQPwCqd;LEVEL-3&filter=pe:2018&**aggregationType=undefined**&skipMeta=false&skipData=true&includeMetadataDetails=true_

This request returns 500 server error.
Removing `aggregationType=undefined` from analytics URL makes request correct and it returns correct response. 

### Proposal:
To fix this we can check if option value (e.g. aggregationType) is undefined and if it is - check if defaultValue is/can be undefined. If the answer to both questions is yes, then assume we can assign undefined to option value, otherwise skip assigning (see files changed).

### DIFF
--- a/packages/plugin/src/ChartPlugin.js
+++ b/packages/plugin/src/ChartPlugin.js
@@ -52,12 +52,9 @@ class ChartPlugin extends Component {
         const options = getOptionsForRequest().reduce(
             (map, [option, props]) => {
                 // only add parameter if value !== default
-                // only assign undefined if defaultValue is/can be undefined
                 if (
-                    (visualization[option] !== undefined &&
-                        visualization[option] !== props.defaultValue) ||
-                    (visualization[option] === undefined &&
-                        visualization[option] === props.defaultValue)
+                    visualization[option] !== undefined &&
+                    visualization[option] !== props.defaultValue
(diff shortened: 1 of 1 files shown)


neeilya:
> Perhaps instead of passing `undefined` in the query string, we can just skip those arguments.

Just to clarify, you suggest to remove/make non-requestable `aggregationType` and `completedOnly` from [options](https://github.com/dhis2/data-visualizer-app/blob/master/packages/plugin/src/modules/options.js#L24-L25) object which we [use to iterate through](https://github.com/dhis2/data-visualizer-app/blob/master/packages/plugin/src/modules/options.js#L53) when we generate options for request?

edoardo:
No, what I mean is to only add `aggregationType` and `completedOnly` to the request arguments only if the value is **not** `undefined` and different from the default.

neeilya:
Thanks, fix pushed.

PR #242, opened 2019-04-09
chore: use @dhis2/d2-ui-analytics package for shared components

jenniferarnesen:
The internal "shared" package is removed. Instead, get ItemSelector from @dhis2/d2-ui-analytics, and move the colors file back to app since it is only used there, and not in the plugin.

This PR does not introduce any functional changes.

The next step is to use the DataDimension and DynamicDimension from @dhis2/d2-ui-analytics, but that will be done as a separate commit.

Test cases:
- check that DataDimension and DynamicDimension selectors work as before
- cursory check that colors in the app are as before.


edoardo:
`DimensionsPanel` and the various dialogs can also be imported from `d2-ui-analytics` now, similarly to how it's done in `dashboards-app`.
This requires some changes to adapt the app's `DialogManager` and possibly other parts.

jenniferarnesen:
> `DimensionsPanel` and the various dialogs can also be imported from `d2-ui-analytics` now, similarly to how it's done in `dashboards-app`.

Yes, I thought I would do that as a separate step. This PR is mostly about removing the "shared" package.

edoardo:
Right. I should have read the PR description better 😉

PR #249, opened 2019-04-10
chore: use specific app class instead of generic 'app' to prevent side effects`

jenniferarnesen:


jenniferarnesen:
Problem with Travis integration - try again with new PR

PR #254, opened 2019-04-12
fix: Use deep equality check for plugin props on componentDidUpdate to avoid unnecessary updates

neeilya:
Fixes [DHIS2-6597](https://jira.dhis2.org/browse/DHIS2-6597).

#### Problem:
Chart items re-render in dashboards app even if props stay the same (triggers by editing text item).

#### Proposal:
Use `lodash/isEqual` function to check equality of props (config & filter) to determine if component should re-render the chart.


neeilya:
Going to merge this one as a temp fix until we fix store design in dashboards app.

PR #257, opened 2019-04-15
Add pointer cursor to recommended icon

janhenrikoverland:


janhenrikoverland:
Moving the fix to d2-ui-analytics

PR #260, opened 2019-04-16
chore: replace built-in dimension panel and dialog components with d2-ui-analytics shared components

edoardo:
Remaining tasks:
* ~decide what to do about the disabled addMetadata call~ (see comments below)


jenniferarnesen:
Comment from Edoardo:

"I had to remove the call to add metadata when the analytics response comes back as it overrides some metadata info that is previously set and required data goes missing (ie. ou path) causing breakage in OrgUnitDimension (at least). For testing you need latest `d2-ui-analytics`"

jenniferarnesen:
Some thoughts:
- what about keeping the original DV DimensionsPanel file and set up all the redux for the panel there? App.js is pretty big already.
- so the options context menu will remain in the apps and not be part of the shared component?

edoardo:
> * in DialogManager, why are the `peIds` and `dxIds` props needed? They seem to be available in the `selectedItems` prop.

Yes, they are and I did consider removing those 2 props, but then I saw they were used also for comparing prevProps with props. We could remove them and use `selectedItems` instead.

packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
```diff
@@ -81,23 +165,53 @@ export class DialogManager extends Component {
...
+        };
+
         const dynamicContent = () => {
+            const selectedItems = this.getSelectedItems(dialogId);
+
             if (dialogId === dxId) {
-                return <DataDimension />;
+                return (
+                    <DataDimension
+                        displayNameProp={displayNameProperty} // XXX from user settings but metadata only has name
```
jenniferarnesen: Is this comment still relevant?

edoardo: No, if `displayName` is available it's always added to the object stored in metadata.

edoardo:
Just to comment on the metadata issue.
I changed the code a bit to use `metadata` and `parentGraphMap` in the Redux store as before.
`metadata` has been optimised to avoid storing information that we don't need.
For orgunits specifically, the path is handled in `parentGraphMap`.
For all dimensions we only store `id`, `name` and `displayName` in `metadata`; the current need is just for the dimension's name/displayName lookup by id.

PR #278, opened 2019-04-29
Publish data-visualizer-plugin with @dhis2 namespace - v32

jenniferarnesen:
We are now publishing @dhis2/data-visualizer-plugin with the @dhis2 namepace


jenniferarnesen:
Why isn't @dhis2/data-visualizer-plugin being added to yarn.lock?

amcgee:
So I believe the issue is that yarn workspaces omits local packages from yarn.lock, so since the package is also in the same workspace it's presumed not to be resolved through npm.

I'm not sure why `d2-charts-api@32.0.1` is being added though, when I run `yarn` on this branch it is removed.

PR #280, opened 2019-04-29
fix: Use deep equality check for plugin props to avoid unnecessary updates - v32

jenniferarnesen:
Fixes DHIS2-6597.

Problem:
Chart items re-render in dashboards app even if props stay the same (triggers by editing text item).

Proposal:
Use lodash/isEqual function to check equality of props (config & filter) to determine if component should re-render the chart.


jenniferarnesen:
This fix causes a different bug

PR #281, opened 2019-04-29
Revert "Use deep equality check for plugin props (#254)"

jenniferarnesen:
This reverts commit 93f71ab9b5a46bb83fce377531137d371cf6a6a6.

This fix causes a different bug - when resizing a chart in edit mode, the chart is not regenerated


jenniferarnesen:
Not reverting after all.

PR #298, opened 2019-06-04
feat: summarize chart filters that have orgunit levels and/or groups

jenniferarnesen:
If a chart filter contains orgUnit levels and or groups, then the filter text will be summarized instead of listing each ou in the level/group. This is to avoid having very long chart filter texts.

Changes include:
- request ou levels once at app startup, and store as prop in the App component
- convert numeric ou ids to uids, which may be used when generating the chart title, as well as correctly populating the orgunit dialog
- update apiVersion to current
- convert args for tDoLoadVisualization to an object, since it had reached 4 args
- remove unused functions: apiFetchOrganisationUnits, apiFetchOrganisationUnitGroups
- moved `getOrgUnitsFromIds` to the DialogManager since it is only relevant for DialogManager

[DHIS2-6990]


packages/app/src/modules/orgUnit.js
```diff
@@ -1,11 +1,73 @@
...
+    const hasNumberedOuLevels =
+        ouPart && ouPart.items.some(item => isNumericOuLevel(item.id));
+
+    if (hasNumberedOuLevels) {
+        const updatedOu = await getUpdatedOuLevels(visualization[prop], ouPart);
+
+        return Object.assign({}, visualization, { [prop]: updatedOu });
+    }
+
+    return visualization;
```
janhenrikoverland: There are so many operations going on in this function. I think we should add a couple of new generic functions to /modules/layout and simplify it.

packages/app/src/modules/orgUnit.js
```diff
@@ -1,11 +1,73 @@
...
+            });
+        }
+        return item;
+    });
+    const newOuFilter = Object.assign({}, ouPart, { items });
+
+    return visualizationProp
+        .filter(p => p.dimension !== DIMENSION_ID_ORGUNIT)
+        .concat(newOuFilter);
+};
```
janhenrikoverland: I suspect that this will change the order of the dimensions in the axis?

packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
```diff
@@ -136,6 +134,18 @@ export class DialogManager extends Component {
...
     };
 
+    getOrgUnitsFromIds = (ids, metadata, parentGraphMap) =>
+        ids
+            .filter(id => metadata[ouIdHelper.removePrefix(id)] !== undefined)
+            .map(id => {
+                const ouUid = ouIdHelper.removePrefix(id);
+                return {
+                    id,
+                    name: metadata[ouUid].displayName || metadata[ouUid].name,
```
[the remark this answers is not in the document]
jenniferarnesen: The key will exist, because the list has already been filtered by existing id (see line 139)

packages/app/src/modules/orgUnit.js
jenniferarnesen: Fixed

packages/app/src/modules/orgUnit.js
```diff
@@ -1,11 +1,73 @@
+import { ouIdHelper, DIMENSION_ID_ORGUNIT } from '@dhis2/d2-ui-analytics';
+
+import { apiFetchOrganisationUnitLevels } from '../api/organisationUnits';
+
+const isNumericOuLevel = id =>
+    ouIdHelper.hasLevelPrefix(id)
+        ? Number.isInteger(parseInt(ouIdHelper.removePrefix(id), 10))
+        : false;
+
+const getUpdatedOuLevels = async (visualizationProp, ouPart) => {
```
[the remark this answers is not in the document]
jenniferarnesen: Updated

packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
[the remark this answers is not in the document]
jenniferarnesen: I think I agree on a metadata helper. After this PR, we can look into all the uses in DV and come up with a suggestion.

packages/app/src/modules/orgUnit.js
jenniferarnesen: Shall we discuss?

PR #300, opened 2019-06-19
task: single value chart type

edoardo:
The PR adds the new visualization type in the type selector and implements the logic for the active/inactive items and the message in the info box in DataDimension shared component.

On a side note, I spoke with @cooper-joe about the possibility of changing the text in the info box to make it clear that is possible to reorder the items even if they look inactive, and that reordering affects the value rendered in the single value visualization.


jenniferarnesen:
Chip color issue in item selector:

When open the facility type item selector, and choose more than one facility type, the chip colors for the non-top items are faded green, but they should be the normal chip green for selected items.  It may be getting this style from the rules for the data element.

image

edoardo:
> Chip color issue in item selector:
> 
> When open the facility type item selector, and choose more than one facility type, the chip colors for the non-top items are faded green, but they should be the normal chip green for selected items. It may be getting this style from the rules for the data element.
> 
> image

Fixed. The active/inactive style was applied also for dynamic dimensions. Now it's only passed to the DataDimension component.

packages/app/src/modules/layoutValidation.js
```diff
@@ -120,6 +121,15 @@ const validatePieLayout = layout => {
     );
 };
 
+const validateSingleValueLayout = layout => {
+    validateAxis(layout.columns, errorLabels.defaultSeries);
+    validateAxis(layout.filters, errorLabels.pie.filter);
+    validateDimension(
+        layoutGetDimension(layout, DIMENSION_ID_PERIOD),
+        errorLabels.pie.pe
```
jenniferarnesen: Confirm that this should be "pie"

edoardo: It could have be if the error message is the same. But I added custom error messages for SV now, which are more appropriate.

PR #316, opened 2019-09-05
fix: error handling in DV app/plugin

edoardo:
Problem was that when changing visualization type and the layout validation failed, the plugin code was still running (due to async) and generated errors due to incompatible data passed in the object.
For example, switching from a gauge/sv AO to column, the layout
validation fails because of missing dimension in category. The error was
correctly raised, but the async code in the plugin was also raising an
error which was overriding the layout error.
The solution is to run the validation before we render the Visualization
component, in App.js. In this way the error is passed directly to the
BlankCanvas component and the Visualization component is never rendered.
Also, the Visualization component does not need to check for errors and
decide if render the plugin or the BlankCanvas component.

### DIFF
--- a/packages/app/src/components/App.js
+++ b/packages/app/src/components/App.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
 import PropTypes from 'prop-types';
 import i18n from '@dhis2/d2-i18n';
 import HeaderBar from '@dhis2/ui/widgets/HeaderBar';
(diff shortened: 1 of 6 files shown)


edoardo:
Continuing the discussion.
The problem with running the validateLayout before the Update button is clicked, is that the current object passed is the original one, not the one reflecting the UI (which changes when vis type is changed).

Another problem happens when switching to certain viz types (ie. SV) because when the Update button is clicked, in order to update current, `getSingleValueCurrentFromUi()` is run, which assumes axes have dimensions/items set, which is not always the case.

edoardo:
Closing in favour of #317 to avoid confusion.

PR #317, opened 2019-09-10
Fix error handling when switching viz type

edoardo:
Problem was that when changing visualization type and the layout validation failed, the plugin code was still running (due to async) and generated errors due to incompatible data passed in the object.
For example, switching from a gauge/sv AO to column, the layout
validation fails because of missing dimension in category. The error was
correctly raised, but the async code in the plugin was also raising an
error which was overriding the layout error.
The solution is to run the validation before we render the Visualization
component, in Layout.js and UpdateButton.js.
In this way the plugin is never rendered, instead 
BlankCanvas component is used to show the error.

Fixed also an issue in the function that extracts current from ui for SV, in some cases the dimensions/items are not populated in ui, ie. when creating a new visualization.


edoardo:
Must be backported to v33.

PR #327, opened 2019-10-03
refactor: port plugin and app to app-platform

amcgee:
This ports the data-visualizer app and plugin to use the app platform.  It gives us a much cleaner package.json and quite a few deleted (unnecessary) files.  The application initialization pieces are also simpler.

~Two features in `app-platform` are currently blocking this port:~
- [x] We don't support named imports from cjs modules in rollup yet, this will be required (the alternative would be to modify a ton of import statements in the codebase here, which is also possible) - see https://github.com/dhis2/app-platform/issues/69
- [x] We don't currently support `import x from 'y.png'` static asset imports yet.  There's only one case of this in data-visualizer, but it would be good to support image loading as a first-class platform feature.

Feel free to take a gander or dig in to help!


packages/app/src/D2Shim.js
```diff
@@ -0,0 +1,64 @@
...
+
+    d2config.baseUrl = `${baseUrl}/api/${apiVersion}`;
+    d2config.schemas = ['chart', 'organisationUnit', 'userGroup'];
+
+    const userSettings = extractUserSettings(await getUserSettings());
+
+    await configI18n(userSettings);
+
+    const d2 = await d2Init({
+        baseUrl: config.baseUrl,
```
jenniferarnesen: `config` doesn't exist, right? (you renamed it to d2Config above)

amcgee: Correct, thanks for the catch!

amcgee: Actually, though, this should just be **baseUrl** - I desctructured the argument which was previously called **config**

amcgee:
@edoardo This should be fully functional, jest and cypress tests also pass.  We need to wait for dhis2/app-platform#187 to be merged before we can merge this, but for now I'm pointing at the pre-release version so everything should work locally!

amcgee:
NOTE: we'll need to replace or update the `deploy-build` script to make this work on master, it won't find the new build output (@varl)

amcgee:
A standalone Netlify deployment is available [here](https://deploy-preview-327--dhis2-data-visualizer.netlify.com/) - you can test against the server `https://debug.dhis2.org/dev`

varl:
> 
> 
> NOTE: we'll need to replace or update the `deploy-build` script to make this work on master, it won't find the new build output (@varl)

Would be nice if we could deprecate the `deploy-build` script by moving the functionality to `d2` CLI. 🤔

It would be easier to configure what build directory to upload in e.g. `d2.config.js`.

packages/app/package.json
```diff
@@ -6,139 +6,54 @@
...
-        "/dhis-web-core-resource": {
-            "target": "http://localhost:8080"
-        }
-    },
-    "babel": {
-        "presets": [
-            "react-app"
+    "jest": {
+        "transformIgnorePatterns": [
+            "node_modules/(?!(lodash-es|analytics-shared|@dhis2/d2-ui-[a-z-]+)/)"
```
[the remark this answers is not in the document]
amcgee: Good catch

amcgee:
> Would be nice if we could deprecate the `deploy-build` script by moving the functionality to `d2` CLI. 🤔

Agreed!  `d2 utils release --publish d2-ci` should be quite possible (just need to write a `d2-ci` plugin for semantic-release)

amcgee:
@varl actually maybe this will work with the third parameter we added - https://github.com/dhis2/usage-analytics-app/blob/master/.travis.yml#L13

amcgee:
@janhenrikoverland @edoardo I had to make some functional changes to get the merge working, it looks like there were some things broken on the epic branch.  Does [this change](https://github.com/dhis2/data-visualizer-app/pull/327/commits/e224251fee48b9424f74c639283b2db7598c36bf) and the [merge commit](https://github.com/dhis2/data-visualizer-app/pull/327/commits/cdacf5025dd1b43d9ed440d04a1a305fdf5712c1) look OK?  If so feel free to approve and merge this branch to the epic branch.

janhenrikoverland:
Looks good. We should import all vis types (previously chart types) from analytics now, but we could refactor that properly in a separate PR.

edoardo:
Yes. I agree with Jan. We can remove chartTypes from the plugin in a separate PR.

PR #362, opened 2019-11-04
feat: use layout rules for add-to-layout ui

janhenrikoverland:
Requires @dhis2/analytics@2.5.0 (https://github.com/dhis2/analytics/pull/130)

- Replaced hardcoded add-to-layout axis logic with layout ui rules
- Fix bug where non-applicable axes were shown as axis options
- Some tests commented out, will fix later: DHIS2-7809


packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/AddToLayoutButton.js
```diff
@@ -62,81 +45,91 @@ export class AddToLayoutButton extends Component {
         this.props.closeDialog(null);
     };
 
+    getAvailableAxisNames = () =>
+        getAvailableAxes(getLayoutTypeByChartType(this.props.ui.type));
+
+    getAxisMeta = axisNameArray =>
```
janhenrikoverland: Use centralized version from the dimension-option PR

PR #371, opened 2019-11-15
fix: map is not vis type

martinkrulltott:


packages/app/src/components/VisualizationTypeSelector/VisualizationTypeMenuItem.js
```diff
@@ -1,38 +1,36 @@
...
             style={styles.listItemText}
         />
     </MenuItem>
 );
 
 VisualizationTypeMenuItem.propTypes = {
-    type: PropTypes.oneOf(Object.keys(visTypeDisplayNames)),
-    visualizationType: PropTypes.oneOf(Object.keys(visTypeDisplayNames)),
     styles: PropTypes.object,
 };
```
edoardo: Add `iconType`, `label`, `isSelected` prop types

edoardo: And what about `onClick` and `disabled`? I see those are passed along to MenuItem, so in that case I'm not sure where the prop type should be added...

martinkrulltott: As they're only getting sent through to the MUI component without any interference prop types would be redundant / unnecessary, since the component would be agnostic to the fact. Otherwise we'd need to specify all MUI props available.

PR #372, opened 2019-11-15
fix: changed incorrect prop types and tests

martinkrulltott:


martinkrulltott:
DONT MERGE, still needs updated version of Analytics

PR #445, opened 2019-11-20
fix: rename axis name constants

janhenrikoverland:
DO NOT MERGE, depends on https://github.com/dhis2/analytics/pull/153

Use renamed constants (axis name -> axis id) from analytics.

### DIFF
--- a/packages/app/src/modules/current.js
+++ b/packages/app/src/modules/current.js
@@ -22,12 +22,15 @@ const hasItems = (object, id) =>
 
 export const getAxesFromUi = ui =>
     Object.entries(ui.layout).reduce(
-        (layout, [axisId, ids]) => ({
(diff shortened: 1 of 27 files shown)


packages/app/src/modules/current.js
```diff
@@ -22,9 +22,9 @@ const hasItems = (object, id) =>
 
 export const getAxesFromUi = ui =>
     Object.entries(ui.layout).reduce(
-        (layout, [axisName, ids]) => ({
+        (layout, [axisId, ids]) => ({
```
martinkrulltott: What's "ids" compared to axisId?

janhenrikoverland: Renamed to "dimensionId".

PR #487, opened 2019-11-25
fix: use layout type to get the right layout component

janhenrikoverland:
Use layout type to get the right layout component.


martinkrulltott:
Superseeded by https://github.com/dhis2/data-visualizer-app/pull/563

PR #491, opened 2019-11-28
fix: locked dims not array

martinkrulltott:
* Updated Analytics dep to v2.6.12
* Locked dimensions handled as object instead of array
* General update to match latest Analytics signatures


packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
```diff
@@ -213,7 +213,7 @@ export class DialogManager extends Component {
             const visType = type;
             const numberOfItems = selectedItems.length;
 
-            const maxNumberOfItemsPerAxis = getMaxNumberOfItemsPerAxis(
+            const maxNumberOfItemsPerAxis = getAxisMaxNumberOfItems(
```
martinkrulltott: Consider changing the const name correspondingly.

packages/app/src/components/Layout/Chip.js
```diff
@@ -44,12 +44,11 @@ class Chip extends React.Component {
...
-    isLocked = getLockedDimensionAxis(
-        this.props.type,
-        this.props.dimensionId
-    ).includes(this.props.axisId);
+    isLocked =
+        getAxisPerLockedDimension(this.props.type, this.props.dimensionId) ===
+        this.props.axisId;
 
-    maxNumberOfItemsPerAxis = getMaxNumberOfItemsPerAxis(
+    maxNumberOfItemsPerAxis = getAxisMaxNumberOfItems(
```
martinkrulltott: maxNumberOfItemsPerAxis => axisMaxNumberOfItems ?

PR #492, opened 2019-11-28
refactor: tooltip

martinkrulltott:


martinkrulltott:
Something went wrong, got lots of commits that weren't intended. Retrying! :)

PR #519, opened 2019-12-06
feat: drag 'n drop for reordering dimension items in layout

edoardo:
This adds drag 'n drop capabilities to the dimension Chip components in
the Layout, allowing for reordering of the dimensions within an axis.
This is required for PT where the order is reflected in the generated
table.

DHIS2-7939.

### DIFF
--- a/packages/app/src/components/Layout/DefaultLayout/DefaultAxis.js
+++ b/packages/app/src/components/Layout/DefaultLayout/DefaultAxis.js
@@ -138,7 +138,7 @@ class Axis extends React.Component {
                                                 {...provided.dragHandleProps}
                                             >
                                                 <Chip
-                                                    key={key}
+                                                    key={`${key}-chip`}
                                                     onClick={this.props.getOpenHandler(
                                                         dimensionId
                                                     )}
--- a/packages/app/src/components/Layout/DefaultLayout/DefaultAxis.js
+++ b/packages/app/src/components/Layout/DefaultLayout/DefaultAxis.js
@@ -105,43 +105,57 @@ class Axis extends React.Component {
 
     render() {
         return (
-            <Droppable droppableId={this.props.axisId} direction="horizontal">
-                {provided => (
-                <div
-                    id={this.props.axisId}
-                    style={{ ...styles.axisContainer, ...this.props.style }}
-                    onDragOver={this.onDragOver}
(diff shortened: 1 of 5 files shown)


packages/app/src/components/Layout/DefaultLayout/DefaultAxis.js
```diff
@@ -111,16 +112,48 @@ class Axis extends React.Component {
...
+                                        index={index}
+                                    >
+                                        {provided => (
+                                            <div
+                                                ref={provided.innerRef}
+                                                {...provided.draggableProps}
+                                                {...provided.dragHandleProps}
+                                            >
+                                                <Chip
+                                                    key={key}
```
martinkrulltott: is it ok to reuse the same key for a different element? append "-chip" or "-child"?

packages/app/src/components/Layout/Layout.js
```diff
@@ -42,10 +45,52 @@ const getLayoutByType = (type, props) => {
     return <Layout {...props} />
 }
 
-const Layout = props => getLayoutByType(props.type)
+const Layout = props => {
+     const onDragEnd = result => {
+        const { source, destination } = result;
+
+        if (!destination) {
```
martinkrulltott: in which scenario is `!destination` true? any case for `!source`?
if so, suggestion: `!destination || !source`

packages/app/src/components/Layout/DefaultLayout/DefaultAxis.js
edoardo: I guess it's ok, but I will add the prefix/postfix.

packages/app/src/components/Layout/Layout.js
edoardo: I think is when you drop a Draggable outside a Droppable, in that case nothing should happen.
The source on the other hand, should always be present at drag end, because it's the Draggable element you are moving.

packages/app/src/components/Layout/DefaultLayout/DefaultAxis.js
amcgee: Chip shouldn’t need a key at all here, right?

edoardo: Right, it shouldn't be needed.

PR #524, opened 2019-12-12
refactor: hide and update buttons

edoardo:
- use shared hide and update buttons.
- fixed MeasureCriteria value handling
- fixed various eslint errors


packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
```diff
@@ -130,6 +130,8 @@ export class DialogManager extends Component {
         }
     }
 
+    onClose = () => this.props.closeDialog(null);
```
martinkrulltott: rename to "closeDialog" or just "close"?

packages/app/src/components/VisualizationOptions/styles/VisualizationOptions.style.js
```diff
@@ -1,26 +1,26 @@
-import React from 'react';
-import css from 'styled-jsx/css';
+import React from 'react' /* eslint-disable-line no-unused-vars */
```
martinkrulltott: why disable the rule?

packages/app/src/modules/options/pivotTableConfig.js
```diff
@@ -32,27 +32,44 @@ export default [
         content: [
             {
                 key: 'data-section-1',
-                content: [<ShowDimensionLabels />, <SkipRounding />],
+                /* eslint-disable react/jsx-key */
```
martinkrulltott: Consider just only adding one line of `eslint-disable react/jsx-key`to the top of the file, since it seems to be used everywhere anyway. Cleaner and less duplication 🙂

packages/app/src/components/VisualizationOptions/styles/VisualizationOptions.style.js
edoardo: See: https://github.com/dhis2/app-platform/issues/213

edoardo: This is most probably temporary, since it seems for apps we are going to use CSS modules.
But I'm waiting for the final decision on that.

packages/app/src/modules/options/pivotTableConfig.js
edoardo: For other sections we are specifying the key here and expect to have it in the component that use these config files, so I went for this approach.

packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
edoardo: Renamed to `onClose` for maintaining the naming conventions.

packages/app/src/modules/options/pivotTableConfig.js
edoardo: Nevermind my comment. Got it. Fixed.

PR #528, opened 2019-12-16
fix: epi curve, update analytics and plugin deps

janhenrikoverland:
Fix epi curve bug by upgrading deps.


janhenrikoverland:
Closing this PR, wrong approach.

PR #537, opened 2019-12-19
refactor: use data query for load/save a visualization

edoardo:
While working on the `legendSet` option component, I realised I need the legend set `displayName` if I want to be able to populate the select without fetching the whole legendSet list.
So I added the nested fields with the `[]` notation and the data was returned correctly by the API, but the problem is that `d2.model` ignores the list of fields for properties that are of type `REFERENCE`.
`legendSet` happens to be one such property, so I was not able to get the `displayName` in the Redux store as `model.toJSON()` would strip some data based on the schema definition.

So I decided to try use `DataEngine` instead, resulting in this PR.

There are a couple of issues to be solved tho (thus the WIP label):
1. what's the best practice when it comes to error handling when `DataQuery` is used in this fashion? (If I pass a wrong id in the URL which results in a 404, the app breaks)
2. the plugin has the same `apiFetchVisualization` function; i'm not 100% sure but it shouldn't be used as now we always pass the whole object (allowing filtering in dashboards among other things). What should we do with that? It shouldn't be duplicated anyway, but moving it to `@dhis2/analytics` seems also not right as it's a specific DV thing
3. if we need to keep the fetch in the plugin, there is no Redux store there, and the plugin are React classes, so I'm not sure how `DataQuery` should be used there, we can of course evaluate bigger changes in the plugin(s) to accommodate `DataQuery`.
4. to completely get rid of `d2` in DV we need to work on a couple of `d2-ui` components where we need to pass `d2`.

@amcgee I'm waiting for your review and comments.


packages/app/src/api/visualization.js
```diff
@@ -1,46 +1,36 @@
...
-export const apiFetchVisualization = (type, id) =>
-    getInstance().then(d2 =>
-        d2.models[type].get(id, {
-            fields: getFieldsStringByType(type),
-        })
-    )
+export const apiFetchVisualization = (dataEngine, type, id) => {
+    const visualizationQuery = {
+        visualization: {
+            resource: `${type}s`, // XXX SUPER UGLY HACK will be fixed when the vis api PR is merged
```
amcgee: Should this maybe be `TODO:` so we can find it later?

packages/app/src/actions/index.js
```diff
@@ -81,7 +84,7 @@ export const tDoLoadVisualization = ({
     }
 
     try {
-        return onSuccess(await apiFetchVisualization(type, id))
+        return onSuccess(await apiFetchVisualization(engine, type, id))
     } catch (error) {
```
amcgee: This should catch the 404 error, does it not work?

packages/app/src/api/visualization.js
edoardo: yes, but it's actually already addressed here: https://github.com/dhis2/data-visualizer-app/pull/527

amcgee:
> 2. the plugin has the same `apiFetchVisualization` function; i'm not 100% sure but it shouldn't be used as now we always pass the whole object (allowing filtering in dashboards among other things). What should we do with that? It shouldn't be duplicated anyway, but moving it to `@dhis2/analytics` seems also not right as it's a specific DV thing
> 
> 3. if we need to keep the fetch in the plugin, there is no Redux store there, and the plugin are React classes, so I'm not sure how `DataQuery` should be used there, we can of course evaluate bigger changes in the plugin(s) to accommodate `DataQuery`.

I think we should be able to use `useDataQuery` in the plugin and share a query object with the app which uses it in redux.  Ideally though we don't need it since the AO should be passed in from Dashboards

packages/app/src/actions/index.js
edoardo: Yes, the problem is with the error passed from the fetch.
The components expect a string, but the error passed is an object.
Fixed.

edoardo:
Issue 1 solved in the app. There was a problem with the handling of the error parameter in the catch statement.
Issues 2 and 3 solved by removing the code for fetching the visualization in the plugin completely.
Ready to be merged.

packages/app/src/api/visualization.js
```diff
@@ -1,46 +1,36 @@
...
 import { getFieldsStringByType } from '../modules/fields'
 
-export const apiFetchVisualization = (type, id) =>
-    getInstance().then(d2 =>
-        d2.models[type].get(id, {
-            fields: getFieldsStringByType(type),
-        })
-    )
+export const apiFetchVisualization = (dataEngine, id) => {
+    const visualizationQuery = {
```
amcgee: To be pedantic, this query should be "static" (constant outside the function) and accept an `id` dynamic variable:

```js
const visualizationQuery = {
  visualization: {
    resource: 'visualizations',
    id: ({ id }) => id,
    params: {
      fields: getFieldsStringByType('visualization'),
    },
  },
}

export const apiFetchVisualization = (dataEngine, id) => {
  return dataEngine.query(visualizationQuery, { variables: { id } })
}
```
This is a pretty minor point, though, since it's more relevant when used in a declarative context (React hooks)

packages/app/src/api/visualization.js
```diff
@@ -1,46 +1,36 @@
...
-            const apiEndpoint = modelDefinition.apiEndpoint
+    return dataEngine.query(visualizationQuery)
+}
 
-            const options = {
-                skipTranslations: true,
-                skipSharing: true,
-            }
+export const apiSaveVisualization = (dataEngine, visualization) => {
+    const mutation = {
```
amcgee: Since this mutation has a dynamic type and ID it can't be extracted, unfortunately, so this is probably the best way to do it here

packages/app/src/api/visualization.js
```diff
@@ -1,46 +1,36 @@
...
-                    `${apiEndpoint}/${visualization.id}${queryString}`,
-                    visualization,
-                    false
-                )
-            } else {
-                return api.post(`${apiEndpoint}${queryString}`, visualization)
-            }
-        })
+    return dataEngine.mutate(mutation, {
+        onError: error => console.error(error),
```
amcgee: This `onError` callback will log the error, but the promise will also still throw - is that what we want?  Should we really be logging to the console here at all?

PR #538, opened 2019-12-20
refactor: components for legendSet and legendDisplayStrategy options

edoardo:
It seems backend support for having multiple legends by data element in a visualization object is not in place.
The implementation in this PR is then the same as the current one used in PT app.
I guess we can merge this version to start with, while the backend support is added.
I'm not sure if that feature/improvement for legendSet is supposed to be in place for 2.34.

Labelling as WIP until I tested it a bit more and I figure out what's the plan for the multiple legend by data element feature/improvement.

Cfr. Joe's sketch: https://sketch.cloud/s/boZO0/a/zOyWPy/play (click Options -> Legend -> Assign a unique legend to data elements).


edoardo:
For now the option works in the same way as the old app.
The custom legend for data item is not yet supported in the backend.
Joe is going to update the sketch for that feature.

PR #547, opened 2020-01-15
feat: show most viewed saved AO (DHIS2-7835)

martinkrulltott:


packages/app/src/components/Visualization/StartScreen.js
```diff
@@ -0,0 +1,82 @@
...
+
+    getFavorites = async () => {
+        const result = await apiFetchFavorites()
+        this.setState({ favorites: result })
+    }
+
+    getContent = () =>
+        this.props.error ? (
+            <div>
+                <img src={chartErrorImg} alt={i18n.t('Chart error')} />
```
edoardo: Change to a more generic `Visualization error` since we have pivot tables types as well.

PR #554, opened 2020-01-20
fix: fixes for DND between axes

edoardo:
When source and destination axis change during a DND move the `acAddUiLayoutDimensions` action is called which runs the layout rules validation.

When source and destination axis remains the same it means it's a reordering and `acSetUiLayout` is used instead (which does not run the layout rules validation).

There is a problem with SV when dragging a dimension in the Series axis, the Data dimension is moved to filters and the dragged dimension drops in Series.
The Data dimension in SV is a locked dimension, so it should not happen.

Requires this PR to be merged first: https://github.com/dhis2/analytics/pull/253

### DIFF
--- a/packages/app/src/components/Visualization/StartScreen.js
+++ b/packages/app/src/components/Visualization/StartScreen.js
@@ -22,7 +22,7 @@ const StartScreen = ({ error, classes }) => {
             setMostViewedVisualizations(result.visualization)
         }
         fetchData(engine)
-    }, [])
+    }, [engine])
 
     const getContent = () =>
         error ? (
--- a/packages/app/src/components/Visualization/StartScreen.js
+++ b/packages/app/src/components/Visualization/StartScreen.js
@@ -22,7 +22,7 @@ const StartScreen = ({ error, classes }) => {
             setMostViewedVisualizations(result.visualization)
         }
         fetchData(engine)
-    }, [engine])
+    }, []) // eslint-ignore-line react-hooks/exhaustive-deps
 
     const getContent = () =>
         error ? (
--- a/packages/app/src/components/Visualization/StartScreen.js
+++ b/packages/app/src/components/Visualization/StartScreen.js
@@ -22,7 +22,7 @@ const StartScreen = ({ error, classes }) => {
             setMostViewedVisualizations(result.visualization)
         }
         fetchData(engine)
-    }, []) // eslint-ignore-line react-hooks/exhaustive-deps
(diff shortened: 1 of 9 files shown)


packages/app/src/modules/layout.js
```diff
@@ -67,11 +69,14 @@ export const getRetransfer = (layout, transfer, visType) => {
         const destinationAxisId = transfer[id]
         const dimensionsAtDestination = layout[destinationAxisId] || []
 
-        const axisIsFull =
-            dimensionsAtDestination.length ===
-            getAxisMaxNumberOfDimensions(visType, destinationAxisId)
-
-        if (axisIsFull) {
+        if (
```
martinkrulltott: is this if statement the same as `canDimensionBeAddedToAxis`?

packages/app/src/components/Visualization/StartScreen.js
```diff
@@ -22,7 +22,7 @@ const StartScreen = ({ error, classes }) => {
             setMostViewedVisualizations(result.visualization)
         }
         fetchData(engine)
-    }, [])
+    }, [engine])
```
martinkrulltott: wouldn't this make the function rerun if `engine` changes?

amcgee: It would.  Engine should be constant, so it shouldn’t be a problem in practice unless someone’s doing something funky with the provider, but you can also add ‘// eslint-ignore-line react-hooks/exhaustive-deps’ to prevent cli-style from complaining

martinkrulltott: `eslint-ignore-line react-hooks/exhaustive-deps` 👍

edoardo: Added the eslint comment.
Now it raises a warning but the commit is allowed.

packages/app/src/modules/layout.js
edoardo: No.
`canDimensionBeAddedToAxis` will return true if the axis is not full, but in that case we don't need a retransfer.
We only need to work out a retransfer if the axis is full.

packages/app/src/components/Visualization/StartScreen.js
amcgee: What's the warning?  I don't think there should be one...

amcgee: Also it doesn't look like this line has changed?

PR #558, opened 2020-01-23
fix: start screen icons

martinkrulltott:
* Added a subsequent API call to fetch the `type` for each visualization and matched that to Analytics `visTypeIcons`.

### DIFF
--- a/packages/app/src/api/visualization.js
+++ b/packages/app/src/api/visualization.js
@@ -10,22 +10,21 @@ const visualizationQuery = {
     },
 }
 
+const visualizationsQuery = {
+    visualization: {
+        resource: 'visualizations',
+        params: ({ params }) => params,
+    },
+}
+
 export const apiFetchVisualization = (dataEngine, id) => {
     return dataEngine.query(visualizationQuery, { variables: { id } })
 }
 
 export const apiFetchVisualizations = (dataEngine, filter, fields) => {
-    const query = {
-        visualization: {
-            resource: 'visualizations',
-            params: {
-                filter: filter,
-                fields: fields,
-            },
-        },
-    }
-
-    return dataEngine.query(query)
+    return dataEngine.query(visualizationsQuery, {
+        variables: { params: { filter, fields } },
+    })
 }
 
 export const apiSaveVisualization = (dataEngine, visualization) => {
--- a/packages/app/src/api/visualization.js
+++ b/packages/app/src/api/visualization.js
@@ -1,17 +1,17 @@
 import { getFieldsStringByType } from '../modules/fields'
 
-export const apiFetchVisualization = (dataEngine, id) => {
-    const query = {
-        visualization: {
-            resource: 'visualizations',
-            id: ({ id }) => id,
-            params: {
-                fields: getFieldsStringByType('visualization'),
-            },
+const visualizationQuery = {
+    visualization: {
+        resource: 'visualizations',
+        id: ({ id }) => id,
+        params: {
+            fields: getFieldsStringByType('visualization'),
         },
-    }
+    },
+}
 
-    return dataEngine.query(query, { variables: { id } })
+export const apiFetchVisualization = (dataEngine, id) => {
+    return dataEngine.query(visualizationQuery, { variables: { id } })
 }
 
 export const apiFetchVisualizations = (dataEngine, filter, fields) => {
--- a/packages/app/src/api/visualization.js
+++ b/packages/app/src/api/visualization.js
@@ -13,7 +13,10 @@ const visualizationQuery = {
 const visualizationsQuery = {
     visualization: {
         resource: 'visualizations',
-        params: ({ params }) => params,
+        params: ({ visualizationIds }) => ({
+            filter: `id:in:[${visualizationIds}]`,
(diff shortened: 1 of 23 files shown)


packages/app/src/api/visualization.js
```diff
@@ -1,17 +1,31 @@
...
-const visualizationQuery = {
-    visualization: {
-        resource: 'visualizations',
-        id: ({ id }) => id,
-        params: {
-            fields: getFieldsStringByType('visualization'),
+export const apiFetchVisualization = (dataEngine, id) => {
+    const query = {
+        visualization: {
+            resource: 'visualizations',
```
edoardo: About this change, see https://github.com/dhis2/data-visualizer-app/pull/537#discussion_r364139186.

martinkrulltott: The _id_ part is still there. But why is it preferable to keep the query as a _const_ instead of in the function? @amcgee

amcgee: It’s a semantic rule we put in place, so that queries can be shared or stored centrally.  It will also help us with caching in the future.  This is modeled after the Apollo / GraphQL approach.  Queries should be static and customized with variables - you will receive a console warning if they are not stable.

amcgee: Additionally, and most importantly, it enforces a 1-to-1 mapping of query to hook. Since the data is only fetched when you first mount the component, having a fully dynamic query doesn’t make sense - changing the query will either do nothing or lead to very strange behavior on a refetch, since the rendering code will likely assume a particular query response structure which would change if you for some reason decided to change the query resource to something else.

amcgee: Looking at this instance specifically it’s actually not a huge issue, though, since this is using the imperative rather than the declarative accessor.

martinkrulltott: I reverted the change for `apiFetchVisualization`. For now `apiFetchVisualizations` is still using a self contained query, but I'd be happy to change it to follow the other example if someone can give me some guidance on how to get the variables (`filter, fields`) to work?

edoardo: Can't you pass all you need in `variables`?

martinkrulltott: Like this? Can't get it to work...

`const visualizationsQuery = {
    visualization: {
        resource: 'visualizations',
        params: {
            filter: ({ filter }) => filter,
            fields: ({ fields }) => fields,
        },
    },
}

export const apiFetchVisualizations = (dataEngine, filter, fields) => {
    return dataEngine.query(visualizationsQuery, {
        variables: { filter, fields },
    })
}`

image

martinkrulltott: Fixed! Could you guys take a quick look and re-review please?

packages/app/src/components/Visualization/StartScreen.js
```diff
@@ -10,18 +10,39 @@ import { withStyles } from '@material-ui/core/styles'
...
+            const mostViewedVisualizationsResult = await apiFetchMostViewedVisualizations(
+                engine,
+                6
+            )
+            const visualizations = mostViewedVisualizationsResult.visualization
+            if (visualizations && visualizations.length) {
+                const visualizationsResult = await apiFetchVisualizations(
+                    engine,
+                    `id:in:[${visualizations.map(vis => vis.id)}]`,
+                    'id,type'
```
amcgee: I think this should be static in the query until it needs to be dynamic (see above), but in any case it's better to pass `fields` as an array than as a string - `['id', 'type']` so that the data engine can de-duplicate requests in the future.

packages/app/src/api/visualization.js
```diff
@@ -10,10 +10,23 @@ const visualizationQuery = {
     },
 }
 
+const visualizationsQuery = {
+    visualization: {
+        resource: 'visualizations',
+        params: ({ params }) => params,
```
amcgee: Somewhat minor but it's better not to expose the implementation details of what and how these params are constructed.  Since this is only used once and the `fields` are well-defined, it seems like they should be static rather than dynamic.  Also, the query should construct the filter param from a list of IDs rather than accepting a raw, pre-constructed filter string.  I.E.

```js
const visualizationsQuery = {
  visualization: {
    resource: 'visualizations',
    params: ({ visualizationIds }) => ({
      filter: `id:in:[${visualizationIds}]`,
      fields: ['id', 'type']
    })
}
```

packages/app/src/api/visualization.js
```diff
@@ -10,10 +10,23 @@ const visualizationQuery = {
...
 export const apiFetchVisualization = (dataEngine, id) => {
     return dataEngine.query(visualizationQuery, { variables: { id } })
 }
 
+export const apiFetchVisualizations = (dataEngine, filter, fields) => {
+    return dataEngine.query(visualizationsQuery, {
+        variables: { params: { filter, fields } },
+    })
+}
+
```
amcgee: With the above change the API of this function would be a bit simpler:

```js
export const apiFetchVisualizations = (dataEngine, visualizationIds) => {
    return dataEngine.query(visualizationsQuery, {
        variables: { visualizationIds },
    })
}
```

PR #574, opened 2020-01-27
feat: use new pivot engine

amcgee:
This is a work-in-progress, but should have basic pivot table functionality from the new PivotTable component and PivotTableEngine in @dhis2/analytics (published as a pre-release version `3.0.0-alpha.0` for now) - see dhis2/analytics#255


packages/plugin/src/PivotPlugin.js
```diff
@@ -1,192 +1,108 @@
...
+    style,
+    onError,
+    onResponsesReceived,
+    id,
+    d2,
+}) => {
+    const [isLoading, setIsLoading] = useState(true)
+    const [data, setData] = useState(null)
+
+    const remappedOptions = {
```
edoardo: I think we can get rid of this, or is the new engine using some Layout stuff from the old PT app?

amcgee: It's not currently reading from the `visualization` object directly, though it could and then we'd have only a couple options left (like the clipping limit)

amcgee: Not using anything from the old app

edoardo: Ok, so no reason for renaming options then.

PR #575, opened 2020-01-28
feat: add dnd from dimensions to specific position in layout axis

jenniferarnesen:
In order to drag/drop a dimension item from the dimension panel to a specific position in a layout axis, both the dimension panel and the layout components need to share the same dnd library (react-beautiful-dnd "rbd").
* The rbd context has been moved up in the hierarchy to App.js, to contain both the dimension panel and layout as children, so they can drag/drop to each other
* DndContext component sets up the rbd context and handles the drop events
* Created custom dnd dimension panel/list/item components to properly configure rbd 
* DimensionsPanel is simpler, as all the item properties have been moved to the DndDimensionList component
* DndDimensionItem sets up a clone item that will stay in the Dimension Panel while the original item is being dragged (due to rbd not supporting this "toolbox" behaviour out-of-the-box)
* styles for the dnd dimension list and panel are copied from the shared components in @dhis2/analytics

Implements [DHIS2-8121]


packages/app/src/components/DimensionsPanel/DndDimensionList.js
```diff
@@ -0,0 +1,135 @@
...
+        return (
+            <DndDimensionItem
+                key={`${SOURCE_DIMENSIONS}-${id}`}
+                {...itemProps}
+            />
+        )
+    }
+
+    render() {
+        const dimensionsList = this.props.dimensions.map((dimension, index) =>
```
edoardo: Can't this be simplified with `this.props.dimensions.filter(filterMatchingDimensions).map( ... )`? and change `filterMatchingDimensions` to just return the dimension or null?

packages/app/src/components/DndContext.js
```diff
@@ -0,0 +1,143 @@
...
+        if (!canDimensionBeAddedToAxis(type, layout[axisId], axisId)) {
+            return
+        }
+
+        const destList = Array.from(layout[axisId])
+
+        destList.splice(index, 0, dimensionId)
+        const reorderedDimensions = { [axisId]: destList }
+
+        this.props.onReorderDimensions({
```
edoardo: IIUC the case here is a dimension dragged from the DimensionPanel to the layout.
I believe this should use the `onAddDimensions` which runs the layout rules validation.

jenniferarnesen:
@edoardo I've added new commits to implement your suggestions, use css modules and add some tests.

PR #582, opened 2020-01-30
chore: small fixes for app-platform production build

amcgee:
- specify coreApp: true in [d2.config.js](https://platform.dhis2.nu/#/config/d2-config-js-reference)
- correct the name of the app icon file so that it is picked up by the platform's manifest generator.


amcgee:
I've pushed another small fix to this branch, correcting the name of the app icon file so that it is correctly picked up by the platform's manifest generator.

PR #584, opened 2020-01-30
fix: remove ripple effect in dimension list when drag starts/ends

jenniferarnesen:
There are 2 separate fixes in this PR:

* Remove "ripple" effect on drag start.
** Add `transform: none !important` to all items not being dragged to remove ripple effect when a drag is started or finished. The ripple effect was solved previously when there was only 1 list, but with 3 separate lists introduced with the Sections, it was necessary to implement differently
** Use a local variable for incrementing the dnd index for dnd to remove the console warning about duplicate, non-sequential indexes. This duplicate, non-sequential indexes was also causing unexpected onDrag movements, similar to the "ripple" effect.
* Show correct dimension icon on the clone (the clone is shown while dragging).
** use the dimension id for the `id` property of DimensionItem. The id prop is not actually used as a css attribute, but is used to determine the icon, and the options click behaviour. Since the clone doesn't get an onOptionsClick, the fact that the id is duplicated doesn't matter. But the actual dimension id is needed to get the correct icon.  You could argue that the property should be named "dimensionId" instead to avoid confusion. That will be more work since it is a breaking change in the analytics library, but happy to discuss.


packages/app/src/components/DimensionsPanel/styles/DndDimensionItem.module.css
```diff
@@ -2,7 +2,8 @@
 This keeps the dimension item clone from causing 
 a ripple effect in the Dimension list during drag
 */
-.dimensionItemClone ~ li {
+
+.notDragging {
     transform: none !important;
```
martinkrulltott: Just a very small note, since this is a very specific use-case, but personally I'm not a huge fan of `!important` statements, is there possibly any way to solve this without it?

packages/app/src/components/DimensionsPanel/DndDimensionItem.js
```diff
@@ -20,6 +20,7 @@ export class DndDimensionItem extends Component {
         } = this.props
 
         const itemCommonProps = {
+            id,
```
edoardo: Wouldn't this result in the same id being used in both the original and the cloned `DimensionItem` elements?

edoardo: Nevermind, I just read your explanation in the description.

packages/app/src/components/DimensionsPanel/styles/DndDimensionItem.module.css
jenniferarnesen: Nobody is a fan of `!important` ;)

Unfortunately, this is one use case where it seems to be needed. This implementation is a "hack" into react-beautiful-dnd to get it to behave in a way that it doesn't do out of the box.

packages/app/src/components/DimensionsPanel/DndDimensionList.js
```diff
@@ -29,11 +31,11 @@ export class DndDimensionList extends Component {
     isLockedDimension = id => this.props.lockedDimensions.includes(id)
     isRecommendedDimension = id => this.props.recommendedIds.includes(id)
 
-    renderItem = ({ id, name }, index) => {
+    renderItem = ({ id, name }) => {
         const itemProps = {
             id,
             name,
-            index,
+            index: this.dndIndex,
```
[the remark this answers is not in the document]
jenniferarnesen: Done.

PR #585, opened 2020-01-31
fix: fix styling issues in options

edoardo:
Based on Joe's feedback.


cooper-joe:
Looking good. A few things I see on the latest:

- Should be dp8 spacing between these number and title inputs:
image

- Still seeing some cut off checkboxes, FF Mac:
image

PR #587, opened 2020-01-31
fix: loading spinner for plugins (DHIS2-8117)

martinkrulltott:
* Moves the loading spinner for the visualization plugins to the wrapper in the app for increased control and flexibility
* New "Plugin Loading" state added to the store
* Triggers the loading spinner when updating a visualization or opening a saved one.
* The loading spinner is then hidden by a callback function `onLoadingComplete` passed down to the plugins.

* **Known issues**: Clicking Update on an unchanged visualization will cause the loading spinner to spin indefinitely.


packages/app/src/components/Visualization/Visualization.js
```diff
@@ -9,11 +9,11 @@ import VisualizationPlugin from '@dhis2/data-visualizer-plugin'
 import { sGetVisualization } from '../../reducers/visualization'
 import { sGetCurrent } from '../../reducers/current'
 import { sGetUiRightSidebarOpen, sGetUiInterpretation } from '../../reducers/ui'
-import { sGetLoadError } from '../../reducers/loader'
+import { sGetLoadError, sGetIsPluginLoading } from '../../reducers/loader'
```
martinkrulltott: I noticed that I'm mixing `IsPluginLoading` and `PluginLoading`, should be consistent, but which is preferred?

PR #617, opened 2020-02-03
fix: convert numeric options to strings to silence ui-core warnings

edoardo:
And convert the values back to numbers before saving the AO.


packages/app/src/modules/options.js
```diff
@@ -119,11 +119,13 @@ export const getOptionsFromVisualization = visualization => {
     }
 
     // cast option values from Number for some options
-    ;['sortOrder', 'topLimit'].forEach(option => {
-        if (Object.prototype.hasOwnProperty.call(visualization, option)) {
-            optionsFromVisualization[option] = String(visualization[option])
+    ;['baseLineValue', 'targetLineValue', 'sortOrder', 'topLimit'].forEach(
+        option => {
+            if (Object.prototype.hasOwnProperty.call(visualization, option)) {
```
amcgee: Why is this not just `if (visualization.hasOwnProperty(option))` ?

packages/app/src/modules/current.js
```diff
@@ -69,11 +69,13 @@ export const getOptionsFromUi = ui => {
     })
 
     // cast option values to Number for some options
-    ;['sortOrder', 'topLimit'].forEach(option => {
-        if (Object.prototype.hasOwnProperty.call(optionsFromUi, option)) {
-            optionsFromUi[option] = Number(optionsFromUi[option])
+    ;['baseLineValue', 'targetLineValue', 'sortOrder', 'topLimit'].forEach(
+        option => {
+            if (Object.prototype.hasOwnProperty.call(optionsFromUi, option)) {
+                optionsFromUi[option] = Number(optionsFromUi[option])
```
amcgee: Could this ever return `NaN`?

packages/app/src/modules/options.js
edoardo: The linter complains about that form.

packages/app/src/modules/current.js
edoardo: It shouldn't as the UI uses numeric input fields.

packages/app/src/modules/options.js
janhenrikoverland: Do we get e.g. https://eslint.org/docs/rules/no-prototype-builtins from cli-style?

PR #623, opened 2020-02-05
fix: dynamic axis names based on vis type

martinkrulltott:
* Implements dynamic axis names based on vis type and layout type, to display the correct axis names for Pivot (_Columns / Rows_ instead of _Series / Category_)

* Dep: https://github.com/dhis2/analytics/pull/284


packages/app/i18n/en.pot
```diff
@@ -689,3 +662,21 @@ msgstr ""
 
 msgid "Quarters per year"
 msgstr ""
+
+msgctxt ""
+"COLUMNS,\n"
+"                    visType\n"
+"                )}."
+msgid ""
+"Add at least one item to ${getAxisNameByVisType(\n"
```
amcgee: This won't work for translations, unfortunately.  The dynamic part won't be translated, first of all, and secondly the function call shouldn't show up in the translation string - you can use [interpolation](https://www.i18next.com/translation-function/interpolation) for that, but again the axis names need to also be translatable.

amcgee: Looks like the names are translated in the `@dhis2/analytics`, so using interpolation should be enough here.  It should be tested to make sure the translations are complete and correct.

martinkrulltott: Oops, not sure why I missed the interpolation here, using it everywhere else 🙈 thanks for spotting it, fixed with the latest commit

packages/app/src/modules/error.js
```diff
@@ -37,21 +39,29 @@ export class EmptyResponseError extends VisualizationError {
...
 }
 
 export class NoCategoryError extends VisualizationError {
-    constructor() {
+    constructor(visType) {
         super(
             EmptyCategory,
-            i18n.t('Category is empty'),
-            i18n.t('Add at least one item to Category.')
+            i18n.t(`{{visType}} is empty`, {
```
amcgee: Nitpick - this variable isn't actually the `visType` (tangent - if it were `visType` I would recommend spelling out `visualizationType` for someone not familiar with the lingo) , it's the `axisName`.  Descriptive variable names help translators get context, though this might be over the heads of many translators anyway.

packages/app/src/components/Layout/YearOverYearLayout/YearOverYearAxis.js
```diff
@@ -16,7 +19,7 @@ const YearOverYearAxis = props => (
         style={{ ...defaultAxisStyles.axisContainer, ...props.style }}
     >
         <div className="label" style={defaultAxisStyles.label}>
-            {getAxisName(props.axisId)}
+            {getAxisNameByLayoutType(props.axisId, LAYOUT_TYPE_YEAR_OVER_YEAR)}
```
amcgee: This may just be my ignorance showing - Does `YEAR_OVER_YEAR` not have a `visType` association?  It seems we should be able to only use `getAxisNameByVisType` (and name it just `getAxisName` to prevent a breaking change) if it did have a `visType`

amcgee:
The interpolation variables should be renamed, and this needs to wait for the changes to https://github.com/dhis2/analytics/pull/284, but otherwise this looks OK from a code perspective to me.

packages/app/src/modules/error.js
martinkrulltott: Yeah, this is a copy/paste error. Fixed. Also changed all `visType` to `visualizationType` for i18n variables.

packages/app/src/components/Layout/YearOverYearLayout/YearOverYearAxis.js
martinkrulltott: Theoretically yes, we can always access the `visType` somehow in the App, however in this case (`YearOverYearAxis.js`) it's not accessible at the moment but could of course be added. Logically though, from `YearOverYearAxis`'s perspective, the axis name should be fetched explicitly for the layout type `LAYOUT_TYPE_YEAR_OVER_YEAR`, not for the "current visualization type".

martinkrulltott: Regarding the breaking change, this was discussed on Slack and @janhenrikoverland insists on this solution, even though it will cause a breaking change subsequently a version bump.

packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/AddToLayoutButton.js
```diff
@@ -21,6 +25,8 @@ export class AddToLayoutButton extends Component {
 
     state = { anchorEl: null }
 
+    labelPrefix = i18n.t('Add to')
```
amcgee: `Add to` might not translate well, it's better to include the interpolation in-line (`Add to {{axisName}}`), maybe we should do that in the two places this is used instead of adding an instance variable?

packages/app/i18n/en.pot
```diff
@@ -459,16 +456,10 @@ msgid ""
 "you may not have access to it."
 msgstr ""
 
-msgid "Series is empty"
+msgid "{{visType}} is empty"
```
amcgee: Looks like these didn't get re-generated?

packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/AddToLayoutButton.js
amcgee: As an example - in some languages, the object might come before the verb.

amcgee: (Yes, I realize this is [very rare](https://en.wikipedia.org/wiki/Object%E2%80%93verb%E2%80%93subject) but there may be different ways to structure imperatives in different languages)

packages/app/i18n/en.pot
martinkrulltott: It's weird how it doesn't regenerate all the time, had to delete the file and rebuild for it to work. Fixed.

packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/AddToLayoutButton.js
martinkrulltott: Indeed, i18n strings should retain as much context as possible. Fixed

packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/AddToLayoutButton.js
```diff
@@ -36,31 +40,27 @@ export class AddToLayoutButton extends Component {
...
+                        ),
+                    })}
                 </MenuItem>
             ))
 
     render() {
-        const availableAxisMeta = this.getAxisMeta(
-            getAvailableAxes(this.props.visType)
-        )
+        const availableAxis = getAvailableAxes(this.props.visType)
```
janhenrikoverland: Should be `const availableAxes` ?

janhenrikoverland: As we're also doing `availableAxis.length`

packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/__tests__/Menu.spec.js
```diff
@@ -2,10 +2,11 @@ import React from 'react'
 import { shallow } from 'enzyme'
 import Menu from '@material-ui/core/Menu'
 import MenuItem from '@material-ui/core/MenuItem'
-import { ADD_TO_LAYOUT_OPTIONS } from '../../../../../modules/layout'
 
 import { DropDown } from '../Menu'
 import DropDownButton from '../DropDownButton'
+import { VIS_TYPE_COLUMN } from '@dhis2/analytics/build/modules/visTypes'
```
janhenrikoverland: Wrong import path

packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/AddToLayoutButton.js
martinkrulltott: Indeed, fixed

packages/app/src/components/DimensionsPanel/Dialogs/AddToLayoutButton/__tests__/Menu.spec.js
martinkrulltott: Indeed, silly VSCode, fixed

packages/app/src/components/Visualization/StartScreen.js
```diff
@@ -67,7 +67,7 @@ const StartScreen = ({ error, classes }) => {
                 {mostViewedVisualizations.length > 0 && (
                     <div style={styles.section}>
                         <h3 style={styles.title}>
-                            Most viewed charts and tables
+                            Your most viewed charts and tables
```
amcgee: This should also be translated with `i18n`!

packages/app/i18n/en.pot
```diff
@@ -34,6 +34,9 @@ msgstr ""
 msgid "Update"
 msgstr ""
 
+msgid "Add to"
```
amcgee: Looks like this needs to be re-generated again, running `yarn build` should do it.

martinkrulltott: Yeah sorry for missing the regeneration again, I can't get it to do it automatically, seems `yarn build` doesn't do it for me. Only way I've found so far is to delete it before the build.

amcgee:
Strange, can you open an issue with repro steps on app-platform?

packages/app/src/components/Visualization/StartScreen.js
martinkrulltott: Indeed. Fixed

PR #624, opened 2020-02-06
fix: download menu options for PT (DHIS2-8207)

edoardo:
Also added missing optional parameters controlled by
options/interpretations.

There are some more parameters that could be passed, looking at this list: https://docs.dhis2.org/master/en/developer/html/webapi_analytics.html#webapi_analytics_query_parameters.
The current implementation in `d2-analysis` seems not all correct.
@janhenrikoverland do we really need support for all those? also the ones that were never handled in the classic apps?


janhenrikoverland:
Let's just replicate what we had in classic PT for now.

edoardo:
@janhenrikoverland this is the same, with the addition of `relativePeriodDate` and without `user` which is not even listed in the available parameters of the analytics request.

See: https://github.com/dhis2/d2-analysis/blob/master/src/api/Layout.js#L991-L994

PR #629, opened 2020-02-07
fix: prevent undefined in adapter

martinkrulltott:
This fix prevents the ghost dimensions (screenshots) that were caused by the new layout adapters recently (https://github.com/dhis2/data-visualizer-app/pull/586/files). 

ghost dimensions


martinkrulltott:
Duplicated by https://github.com/dhis2/data-visualizer-app/pull/630/

PR #636, opened 2020-02-10
fix: remove custom title when Auto is selected (DHIS2-8252)

edoardo:
Also fixed the subtitle option to work in the same way as the title.
Before when no custom subtitle was specified, the Custom subtitle was
selected with an empty value.
Now Auto means not hidden subtitle but also not custom subtitle.
The subtitle is actually auto generated in some situations (depending on
vis type, presence of title...)

### DIFF
--- a/packages/app/src/components/VisualizationOptions/Options/HideSubtitle.js
+++ b/packages/app/src/components/VisualizationOptions/Options/HideSubtitle.js
@@ -17,22 +17,27 @@ import {
     tabSectionOptionToggleable,
 } from '../styles/VisualizationOptions.style.js'
 
+const HIDE_SUBTITLE_AUTO = 'AUTO'
+const HIDE_SUBTITLE_NONE = 'NONE'
+const HIDE_SUBTITLE_CUSTOM = 'CUSTOM'
+
 class HideSubtitle extends Component {
     constructor(props) {
         super(props)
 
-        this.defaultState = { value: 'AUTO' }
+        this.defaultState = { value: HIDE_SUBTITLE_AUTO }
 
         this.state = props.value ? { value: props.value } : this.defaultState
     }
 
     onChange = ({ value }) => {
         this.setState({ value })
-        this.props.onChange({
-            hideSubtitle: value === 'NONE',
-            subtitle:
-                value === 'AUTO' ? undefined : this.props.subtitle || undefined,
-        })
+        this.props.onChange(
+            value === HIDE_SUBTITLE_NONE,
+            value === HIDE_SUBTITLE_AUTO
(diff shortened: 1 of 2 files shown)


packages/app/src/components/VisualizationOptions/Options/HideTitle.js
```diff
@@ -83,10 +87,11 @@ const hideTitleSelector = createSelector([sGetUiOptions], uiOptions =>
 const mapStateToProps = state => ({
     visualizationType: sGetUiType(state),
     value: hideTitleSelector(state),
+    title: sGetUiOptions(state).title,
 })
 
 const mapDispatchToProps = dispatch => ({
-    onChange: enabled => dispatch(acSetUiOptions({ hideTitle: enabled })),
+    onChange: value => dispatch(acSetUiOptions(value)),
```
martinkrulltott: I would prefer if this dispatch function took in `hideTitle` and `title` and constructed it's own parameter object `{hideTitle: X, title, Y}` instead of a pre-constructed object `value`, as it's currently exposing the inner structure of `acSetUiOptions` to the class.

packages/app/src/components/VisualizationOptions/Options/HideTitle.js
```diff
@@ -26,9 +26,12 @@ class HideTitle extends Component {
         this.state = props.value ? { value: props.value } : this.defaultState
     }
 
-    onChange = ({ value }) => {
+    onRadioGroupChange = ({ value }) => {
         this.setState({ value })
-        this.props.onChange(value === 'NONE')
+        this.props.onChange({
+            hideTitle: value === 'NONE',
```
martinkrulltott: Can all of these strings (`NONE`, `AUTO` etc) be standardised as `const` instead?

packages/app/src/components/VisualizationOptions/Options/HideSubtitle.js
```diff
@@ -65,18 +71,28 @@ class HideSubtitle extends Component {
...
 const mapStateToProps = state => ({
     visualizationType: sGetUiType(state),
-    value: sGetUiOptions(state).hideSubtitle ? 'NONE' : 'CUSTOM',
+    value: hideSubtitleSelector(state),
+    subtitle: sGetUiOptions(state).subtitle,
 })
 
 const mapDispatchToProps = dispatch => ({
-    onChange: enabled => dispatch(acSetUiOptions({ hideSubtitle: enabled })),
+    onChange: value => dispatch(acSetUiOptions(value)),
```
martinkrulltott: Same issue/suggestion as in HideTitle.js

PR #637, opened 2020-02-10
fix: chart plugin infinite loading when no change has been made

martinkrulltott:
There's currently a bug with the chart plugin, where it's spinning indefinitely if no change has been made and the Update button is clicked.


martinkrulltott:
This solution causes the loading spinner to not show when opening through the Open menu option, which defeats the whole purpose of the external loading spinner. Back to the drawing board...

martinkrulltott:
Obsolete. Fixed by https://github.com/dhis2/data-visualizer-app/pull/653 instead.

PR #638, opened 2020-02-10
chore(github-actions): add workflows for lint and publish

varl:
We will use Travis for running the deploy-build script to d2-ci ([more information on why](https://developers.dhis2.org/2019/02/the-build-system/#d2-ci-organisation)) and our more _modern_ tooling for NPM until we've ported the deploy-build script.

### DIFF
--- a/.github/workflows/node-publish.yml
+++ b/.github/workflows/node-publish.yml
@@ -1,0 +1,41 @@
+name: 'DHIS2: Release'
+
+on:
+    push:
+        branches:
+            - master
+
+jobs:
+    process:
+        name: Publish
+        runs-on: ubuntu-latest
+        if: "!contains(github.event.head_commit.message, '[skip ci]')"
+        steps:
+            - uses: actions/checkout@v1
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: 12.x
+
+            - name: Install
+              run: yarn install --frozen-lockfile
+
+            - name: Lint
+              run: yarn lint
+
+            - name: Test
+              run: yarn test
+
+            - name: Build
+              run: yarn build
+
+            - name: Publish to NPM
+              run: npx @dhis2/cli-utils release --publish npm
+              env:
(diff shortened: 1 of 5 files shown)


.github/workflows/node-publish.yml
```diff
@@ -0,0 +1,41 @@
...
+                  node-version: 12.x
+
+            - name: Install
+              run: yarn install --frozen-lockfile
+
+            - name: Lint
+              run: yarn lint
+
+            - name: Test
+              run: yarn test
```
amcgee: We're duplicating our testing and linting here... any reason for that?

amcgee:
**NOTE: WE NEED TO PUSH A VERSION TAG BEFORE MERGING THIS PR**

.github/workflows/node-publish.yml
varl: The other workflows for testing and linting run in parallel, so unless we have them here, theoretically we might publish something that fails tests and linting.

We could remove them here if it's too paranoid.

amcgee:
I was more going to suggest removing the other workflows?  You’ll get reports of linting and test failures here...

varl:
> I was more going to suggest removing the other workflows? You’ll get reports of linting and test failures here...

The publish workflow is configured to only run on the master branch.

It's nice to keep them separate. We could turn off lint and test for master branch so only `node-publish` runs on master, but on PRs `node-lint` and `node-test` are run. That way everything only runs once per branch at least.

amcgee:
I'm fine with this for now, no huge harm in duplication here!

I think we should be good to go here, we'll just need to publish a **BREAKING CHANGE** when we want to cut `34.0.0`?

varl:
Yeah, when squashing we need to add that along with a description.

Would this suffice?

`BREAKING CHANGE: Ensure that the plugin and app versions are locked to each other.`

amcgee:
Yeah, sounds good to me

PR #639, opened 2020-02-10
refactor: rename visConfig and config

martinkrulltott:
Currently there's a prop on `Visualization.js` called `visConfig`, which is passed to the plugins as `config`. This name is a legacy from https://github.com/dhis2/data-visualizer-app/commit/af1d0124b0338f0114b9a59fbbc2e500175ba227#diff-e4a1f1987c1f3a83a3ed17722cecb3a4 but has later evolved to contain either `state.visualization` or `state.current`. This PR renames the prop to a more understandable name (currently `visualization`, but that's open for suggestions).

### DIFF
--- a/packages/app/src/components/Visualization/Visualization.js
+++ b/packages/app/src/components/Visualization/Visualization.js
@@ -100,10 +100,10 @@ export class Visualization extends Component {
     }
 
     render() {
-        const { visConfig, visFilters, error } = this.props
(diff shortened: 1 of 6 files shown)


edoardo:
`config` is what was originally called in the old plugins, so I kept that name.
That said I'm fine changing it to `visualization`.
The `visFilter` part can be removed, I don't think we are using it anywhere.

PR #653, opened 2020-02-11
fix: avoid infinite loading and make chart plugin use hooks (DHIS2-8290)

amcgee:
This solves https://jira.dhis2.org/browse/DHIS2-8290 and supercedes #637 - it should work for no-op updates as well as loading saved AOs, changing chart types, and resizing the browser window.

More eyes (on code and testing the app) would be great - it should be functionally equivalent to the previous class Component, but a bit more explicit and clear about what happens when, and with the above bugs fixed, but it's still a fair number of changed lines

### DIFF
--- a/packages/plugin/src/__tests__/ChartPlugin.spec.js
+++ b/packages/plugin/src/__tests__/ChartPlugin.spec.js
@@ -113,7 +113,6 @@ describe('ChartPlugin', () => {
         if (!chartPlugin) {
             chartPlugin = mount(<ChartPlugin {...props} />)
         }
-        chartPlugin
         return chartPlugin
     }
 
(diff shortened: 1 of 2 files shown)


packages/plugin/src/__tests__/ChartPlugin.spec.js
```diff
@@ -108,12 +108,13 @@ describe('ChartPlugin', () => {
         ['option2', { defaultValue: null }],
     ]
     let props
-    let shallowChartPlugin
+    let chartPlugin
     const canvas = () => {
-        if (!shallowChartPlugin) {
-            shallowChartPlugin = shallow(<ChartPlugin {...props} />)
+        if (!chartPlugin) {
+            chartPlugin = mount(<ChartPlugin {...props} />)
```
amcgee: This is necessary because `jest` and `enzyme` [don't run `useEffect` hooks on a shallow mount](https://airbnb.io/enzyme/#react-hooks-support)

PR #655, opened 2020-02-12
fix: dimensions modal to show only when dimension is added

martinkrulltott:
* Prevents an empty modal to be opened when adding Assigned Categories (screenshot)
* Triggers the modal exclusively for _adding_ empty dimensions using both context menus (not when _moving_ as before)

*  **Analytics breaking change**: As `axisItemHandler` requires `requireItems` and `isDimensionInLayout` to implement the behaviour above, the function signature had to be changed to receive an object with 5 props instead of just 3 regular props (because of `eslint max-params`. As seen here: [commit a82b2a7](https://github.com/dhis2/data-visualizer-app/pull/655/commits/a82b2a767ffcc289444505111dd1bde16e53b6dc)

Depends on https://github.com/dhis2/analytics/pull/294

Empty modal that was removed, was previously showing when adding Assigned Categories
image

Move, shouldnt trigger a modal for empty dims
image

Add, should trigger a modal for empty dims
image


martinkrulltott:
Refactor to prevent a breaking change / version bump in Analytics?

packages/app/src/components/DimensionsPanel/DimensionsPanel.js
```diff
@@ -102,10 +102,15 @@ const mapStateToProps = state => ({
 const mapDispatchToProps = dispatch => ({
     dualAxisItemHandler: () =>
         dispatch(acSetUiActiveModalDialog(AXIS_SETUP_DIALOG_ID)),
-    axisItemHandler: (dimensionId, axisId, numberOfDimensionItems) => {
+    axisItemHandler: ({
```
martinkrulltott: Refactor to prevent a breaking change / version bump in Analytics?

packages/app/src/components/Layout/Menu.js
```diff
@@ -94,8 +94,17 @@ const mapStateToProps = state => {
...
-    axisItemHandler: (dimensionId, axisId) => {
+    axisItemHandler: ({
+        dimensionId,
+        axisId,
+        numberOfDimensionItems,
+        requireItems,
+    }) => {
         dispatch(acAddUiLayoutDimensions({ [dimensionId]: { axisId } }))
+
+        if (numberOfDimensionItems === 0 && requireItems) {
```
martinkrulltott: Remove this since it can only "move" dimensions

packages/app/src/components/DimensionsPanel/DimensionsPanel.js
```diff
@@ -102,10 +102,15 @@ const mapStateToProps = state => ({
...
+        dimensionId,
+        axisId,
+        numberOfDimensionItems,
+        requireItems,
+    }) => {
         dispatch(acAddUiLayoutDimensions({ [dimensionId]: { axisId } }))
 
-        if (numberOfDimensionItems === 0) {
+        if (numberOfDimensionItems === 0 && requireItems) {
```
martinkrulltott: Make sure this only triggers on "Add"

PR #657, opened 2020-02-12
feat: option for data approval level

edoardo:
This is a special option, only shown if the user has authority to see unapproved data and the analytics system setting "Max number of years to hide unapproved data in analytics" is set.

In that case the `approvalLevel` with the value set via the UI should be sent as query parameter in the analytics request.


packages/app/src/actions/user.js
```diff
@@ -1,6 +1,30 @@
...
+    engine
+) => {
+    const onSuccess = response => {
+        dispatch(acSetUserAuthority({ [authorityKey]: response.authority }))
+    }
+
+    try {
+        return onSuccess(await apiFetchUserAuthority(engine, authorityKey))
+    } catch (err) {
+        onError('tLoadUserAuthority', err)
```
martinkrulltott: What sort of error does this throw and should it be covered by our [error module](https://github.com/dhis2/data-visualizer-app/blob/master/packages/app/src/modules/error.js)?

packages/app/src/components/App.js
```diff
@@ -114,6 +114,9 @@ export class App extends Component {
             fromActions.fromSettings.tAddSettings(userSettings)
         )
         store.dispatch(fromActions.fromUser.acReceivedUser(d2.currentUser))
+        store.dispatch(
+            fromActions.fromUser.tLoadUserAuthority('F_VIEW_UNAPPROVED_DATA')
```
martinkrulltott: `const` rather than string?

edoardo:
In order to test the feature, a analytic system setting must be changed:
Screenshot 2020-02-13 at 09 28 51

With that, and if the user has the `F_VIEW_UNAPPROVED_DATA` authority set to true, the option becomes available for PT: 
Screenshot 2020-02-13 at 09 30 31

packages/app/src/components/App.js
edoardo: Defined the const in the option component and imported in App.js.

packages/app/src/actions/user.js
edoardo: The fetch is done via the data engine, i guess generic errors can happen.
Not sure there are specific errors for that request, that would be a question for backend.

martinkrulltott: Ah ok, actually I wasn't referring to using a specific error (which would also be good of course), but if this path would trigger `acSetLoadError()`? If not you can catch it and use `acSetLoadError(new GenericServerError())` to display a user friendly error.

edoardo: ok, done.

packages/app/src/components/VisualizationOptions/Options/ApprovalLevel.js
```diff
@@ -0,0 +1,153 @@
...
+            setOptions([
+                ...options,
+                { value: value.id, label: value.displayName },
+            ])
+        }
+    }
+
+    const onSelectFocus = async () => {
+        if (!isLoaded) {
+            const query = {
```
martinkrulltott: Approving this for now, but imho we should add a separate PR to move this query to `src/api` and make it static. Together with this as well https://github.com/dhis2/data-visualizer-app/blob/b602425c10d779b44a4db7d679b521c85e445fa3/packages/app/src/components/VisualizationOptions/Options/LegendSet.js#L83

packages/app/src/components/VisualizationOptions/Options/ApprovalLevel.js
```diff
@@ -0,0 +1,153 @@
...
+                { value: value.id, label: value.displayName },
+            ])
+        }
+    }
+
+    if (!enabled) {
+        return null
+    }
+
+    if (value && value.id) {
```
martinkrulltott: Looks like a copy/paste duplicate of a few lines above

packages/app/src/components/VisualizationOptions/Options/ApprovalLevel.js
edoardo: Moved the query constant at the top as per @amcgee suggestion.

packages/app/src/components/VisualizationOptions/Options/ApprovalLevel.js
edoardo: Fixed

PR #658, opened 2020-02-12
fix: data with multiple indicators error message

martinkrulltott:
New custom error message for when a data indicator is used in combination with other data items.

image

Will be shown when `Data` is in `Filter` and contains an indicator + another item, e.g. 2 indicators or 1 indicator / 1 element. 

Example:

Two indicators
image

Data in filter
image


jenniferarnesen:
@martinkrulltott When you squash merge, could you expand a little on the title/description so it is clear what the commit is about? (sometimes the individual commit messages don't add up to a good explanation)

PR #661, opened 2020-02-12
chore: fix GitHub Actions test workflow

amcgee:
We need to build the plugin before testing, otherwise the plugin's entrypoint (`build/es/lib.js`) doesn't exist and can't be resolved through the workspaces link.

This also removes the extraneous `lerna.json`

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -1,7 +1,6 @@
 name: 'DHIS2: Tests'
 
-on:
-    pull_request:
+on: pull_request
 
 jobs:
     pr:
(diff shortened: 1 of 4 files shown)


.github/workflows/node-test.yml
```diff
@@ -20,6 +17,9 @@ jobs:
             - name: Install
               run: yarn install --frozen-lockfile
 
+            - name: Build
+              run: yarn build
```
varl: I guess this could technically be `yarn build-plugin` when just running the tests on a PR?

For the `node-publish.yml` workflow we def. want to build the entire thing.

amcgee: I think we want to make sure the app production build also succeeds on a PR, we could technically have a separate `build` workflow but this is probably sufficient for now?

varl: Yep, go for it.

PR #662, opened 2020-02-12
fix: correctly specify runtime and peer dependencies

amcgee:
This moves plugin `react` and `react-dom` deps to Peer, and moves app deps from `peer` to `runtime` (technically they could be omitted, but we're going for explicit here).  While it shouldn't technically be required to specify `react` as a peer dep in the plugin (`yarn` should de-duplicate if the major version matches) it makes sense to ensure that there can only be one copy.  This will also cut version 34.0.0.

BREAKING CHANGE: The Data Visualizer Plugin now requires `react` and `react-dom` to be provided by the consumer.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -30,9 +30,7 @@
         "redux-logger": "^3.0.6",
         "redux-thunk": "^2.2.0",
         "reselect": "^4.0.0",
-        "whatwg-fetch": "^3.0.0"
-    },
-    "peerDependencies": {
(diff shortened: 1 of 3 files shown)


amcgee:
I opted to defer the versioning of deps to app-platform, but we could also explicitly specify a dep version here so long as it's within the same major.  I.E. @edoardo you could specify a specific version for `@dhis2/ui-core` when it's needed.

PR #665, opened 2020-02-13
fix: missing Pivot options

martinkrulltott:
* Adds "Display organisation unit hierarchy" to the Style tab of the Options dialog for Pivot
image

* Adds "Only include completed events" to the Data tab of the Options dialog for Pivot
image


martinkrulltott:
> LGTM.
> @cooper-joe I understand _Labels_ is part of this option, not intended as a section in the tab, correct?
> If it's correct, please approve this PR.

@cooper-joe confirmed that this is supposed to be a section. PR updated

edoardo:
@martinkrulltott what about *Event data*?

PR #666, opened 2020-02-13
feat: fetch legendSet and pass as prop to plugins

edoardo:


packages/plugin/src/index.js
```diff
@@ -1,17 +1,42 @@
...
+
     if (
         !props.visualization.type ||
         props.visualization.type === VIS_TYPE_PIVOT_TABLE
     ) {
-        return <PivotPlugin {...props} />
+        return <PivotPlugin {...props} legendSet={legendSet} />
     } else {
-        return <ChartPlugin {...props} />
+        return <ChartPlugin {...props} legendSet={legendSet} />
```
amcgee: This will render the table or chart without the legend first (while it's loading) and then re-render it with the legend once the legendSet loads, is that the intent here?  One option to avoid this would be - 

```js
const VisualizationPlugin = props => {
    const [legendSet, setLegendSet] = useState(null)

    const hasLegendSet = props.visualization.legendSet && props.visualization.legendSet.id

    useEffect(() => {
        const fetchLegendSet = async engine => {
            if (hasLegendSet) {
                const response = await apiFetchLegendSet(
                    engine,
                    props.visualization.legendSet.id
                )

                if (response && response.legendSet) {
                    setLegendSet(response.legendSet)
                }
            }
        }

        fetchLegendSet(engine)
    }, [props.visualization.legendSet]) // eslint-disable-line react-hooks/exhaustive-deps

    if (hasLegendSet && !legendSet) {
        // Until one of the children is rendered and calls onLoadingComplete,
        //   the app will continue to render the loading spinner
        return null
    }

    if(
        !props.visualization.type ||
        props.visualization.type === VIS_TYPE_PIVOT_TABLE
    ) {	
        return <PivotPlugin {...props} legendSet={legendSet} />
    } else {	
        return <ChartPlugin {...props} legendSet={legendSet} />
    }
}

Comment (edoardo) [reply packages/plugin/src/index.js]: Yes, good point.
If the legendSet option is set, we know it before the first render happens, so no point in re-render.

File: packages/plugin/src/index.js [hunk 2]
```diff
@@ -1,17 +1,48 @@
...
 const VisualizationPlugin = props => {
+    const engine = useDataEngine()
+    const [legendSet, setLegendSet] = useState(null)
+
+    const hasLegendSet =
+        props.visualization.legendSet && props.visualization.legendSet.id
+
+    useEffect(() => {
+        const fetchLegendSet = async engine => {
+            if (hasLegendSet) {
```
amcgee: I put this in my sample code but I think we should probably be explicit and re-do the conditional here - otherwise there's an implicit dependency on `hasLegendSet`.  It *should* be fine for now since `hasLegendSet` is just a wrapper around on `props.visualization.legendSet` which is a dependency of the `useEffect`, but it would be easy for us to forget about this and break the effect later.  So I think we should probably just keep this as `props.visualization.legendSet && props.visualization.legendSet.id`, or add `hasLegendSet` to the dep list.

packages/plugin/src/index.js
```diff
@@ -1,17 +1,48 @@
...
+                )
+
+                if (response && response.legendSet) {
+                    setLegendSet(response.legendSet)
+                }
+            }
+        }
+
+        fetchLegendSet(engine)
+    }, [props.visualization.legendSet]) // eslint-disable-line react-hooks/exhaustive-deps
```
amcgee: I don't think we need to disable `eslint` here right?  The only dependencies are `props.visualization.legendSet`, `engine` (which we should add since it's constant), and maybe `hasLegendSet`.  Is there a reason not to list them all?

edoardo: No, I don't see any reason.
The eslint rule is probably a leftover from the start.

packages/plugin/src/index.js
edoardo: ok, I'll revert.

PR #668, opened 2020-02-14
fix: fix default state for LegendSet option

edoardo:
The selection for Legend type was lost when switching between tabs,
causing the app to crash.
The reason was that the radio group had no default value (which does not make sense for a radio group).
`FIXED` should be the default value for `legendDisplayStrategy`, which enables the dropdown for the selection of the legendSet.

Use a constant for the option name and get the default value from the
options list.
Moved also the query to the top.

### DIFF
--- a/packages/app/src/components/VisualizationOptions/Options/LegendSet.js
+++ b/packages/app/src/components/VisualizationOptions/Options/LegendSet.js
@@ -18,11 +18,27 @@ import LegendDisplayStrategy from './LegendDisplayStrategy'
 import { sGetUiOptions } from '../../../reducers/ui'
 import { acSetUiOptions } from '../../../actions/ui'
 
+import { options } from '../../../modules/options'
+
 import {
     tabSection,
     tabSectionOption,
     tabSectionTitle,
 } from '../styles/VisualizationOptions.style.js'
+
+const optionName = 'legendSet'
+const defaultValue = options[optionName].defaultValue
+
+const query = {
+    legendSets: {
+        resource: 'legendSets.json',
+        params: {
+            fields:
+                'id,displayName~rename(name),legends[id,displayName~rename(name),startValue,endValue,color]',
+            paging: false,
+        },
+    },
+}
 
 const LegendSelect = ({ value, loading, options, onFocus, onChange }) => {
     const selected =
@@ -41,10 +57,8 @@ const LegendSelect = ({ value, loading, options, onFocus, onChange }) => {
             onFocus={onFocus}
             onChange={({ selected }) =>
                 onChange({
-                    legendSet: {
-                        id: selected.value,
-                        displayName: selected.label,
-                    },
+                    id: selected.value,
+                    displayName: selected.label,
                 })
             }
         >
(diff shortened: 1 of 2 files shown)


packages/app/src/components/VisualizationOptions/Options/LegendSet.js
```diff
@@ -18,12 +18,28 @@ import LegendDisplayStrategy from './LegendDisplayStrategy'
...
 
+import { options } from '../../../modules/options'
+
 import {
     tabSection,
     tabSectionOption,
     tabSectionTitle,
 } from '../styles/VisualizationOptions.style.js'
 
+const optionName = 'legendSet'
```
martinkrulltott: string consts as all caps with underscore? e.g. `LEGEND_SET_OPTION_NAME`?

packages/app/src/modules/options.js
```diff
@@ -39,7 +39,7 @@ export const options = {
     numberType: { defaultValue: 'VALUE', requestable: false },
     showHierarchy: { defaultValue: false, requestable: true },
     legendSet: { defaultValue: undefined, requestable: false },
-    legendDisplayStrategy: { defaultValue: undefined, requestable: false },
+    legendDisplayStrategy: { defaultValue: 'FIXED', requestable: false },
```
martinkrulltott: What does `legendDisplayStrategy` have to do with this PR? Could you add a comment in the description please

packages/app/src/components/VisualizationOptions/Options/LegendSet.js
```diff
@@ -18,12 +18,28 @@ import LegendDisplayStrategy from './LegendDisplayStrategy'
...
+const defaultValue = options[optionName].defaultValue
+
+const query = {
+    legendSets: {
+        resource: 'legendSets.json',
```
amcgee: The .json extension should not be necessary

packages/app/src/components/VisualizationOptions/Options/LegendSet.js
```diff
@@ -18,12 +18,28 @@ import LegendDisplayStrategy from './LegendDisplayStrategy'
...
+        params: {
+            fields:
+                'id,displayName~rename(name),legends[id,displayName~rename(name),startValue,endValue,color]',
```
amcgee: If it’s helpful or cleaner, this can be specified as an array rather than a string

packages/app/src/modules/options.js
edoardo: That's the actual fix for the original problem.

The LegendSet option is a "advanced" option which uses several simple options: `legendDisplayStrategy`, `legendDisplayStyle` in order to control the value of `legendSet`.
When you choose `FIXED` for `legendDisplayStrategy` the dropdown with the available legends becomes available.

packages/app/src/components/VisualizationOptions/Options/LegendSet.js
edoardo: Fixed

packages/app/src/components/VisualizationOptions/Options/LegendSet.js
edoardo: Fixed

packages/app/src/components/VisualizationOptions/Options/LegendSet.js
edoardo: Done

PR #675, opened 2020-02-18
fix: upgrade analytics dependency

jenniferarnesen:


jenniferarnesen:
Merged in a different PR

PR #696, opened 2020-02-19
refactor: implement latest design for legend option

edoardo:
Design reference: https://sketch.cloud/s/4kvRo/a/y45lJz/play

For 2.34 we don't have support for the last option under "Legend Type", so this PR only implements the other 2.

Because we don't have an option to control the state of the "Display legend" checkbox, its value is computed from the values of the "real" options `legendSet` and `legendDisplayStrategy`.
This means that the old "None" option corresponds to `legendDisplayStrategy` = `FIXED` and `legendSet` not defined, and it renders the checkbox disabled.
This also means that if you select "Select a single legend for entire table" but you don't choose a legend from the dropdown, when you switch tab and come back to the Legend tab, the checkbox will be disabled.


janhenrikoverland:
@edoardo More space between "Display legend" and "Legend style" heading.

PR #697, opened 2020-02-19
refactor: move plugin data fetch to VisualizationPlugin to reduce duplication

amcgee:
This also deletes some duplicated code between app and plugin, and sets the stage for fetching multiple legendSets when `BY_DATA_ITEM` is selected as the legend strategy.

Duplicated code from `ChartPlugin` and `PivotsPlugin` moved to the helper files `getRequestOptions.js` and `fetchData.js`.  The wrapper plugin (`VisualizationPlugin`) now handles all data fetching and just passes the results (`responses`, `extraOptions`, and `legendSets`) to the typed plugin handler.


packages/plugin/src/ChartPlugin.js
```diff
@@ -1,103 +1,21 @@
...
-}
+import { isSingleValue, createVisualization } from '@dhis2/analytics'
 
 const ChartPlugin = ({
     visualization,
-    filters,
-    id,
+    responses,
+    extraOptions,
+    // legendSets,
```
edoardo: We should enable this already as @martinkrulltott is implementing legend set for Gauge.

packages/plugin/src/ChartPlugin.js
```diff
@@ -122,76 +40,41 @@ const ChartPlugin = ({
...
-    onError: Function.prototype,
     onChartGenerated: Function.prototype,
-    onLoadingComplete: Function.prototype,
-    onResponsesReceived: Function.prototype,
 }
 
 ChartPlugin.propTypes = {
-    d2: PropTypes.object.isRequired,
+    extraOptions: PropTypes.object.isRequired,
+    // legendSets: PropTypes.arrayOf(PropTypes.object).isRequired,
```
edoardo: Uncomment this. See above.

packages/plugin/src/modules/getRequestOptions.js
```diff
@@ -0,0 +1,32 @@
+import { getOptionsForRequest } from './options'
```
edoardo: `options.js` is still duplicated between app and plugin.
Perhaps out of scope of this PR, but we should remove that duplication too.
We could import it in the app from the plugin, although it feels it belongs more in the app than in the plugin, given the options UI etc...

packages/plugin/src/ChartPlugin.js
amcgee: @martinkrulltott you can uncomment this line and the one in propTypes when we use legends in ChartPlugin, it is passed by the caller.  It breaks the `no-unused-variables` ESLint rule to leave it here now

packages/plugin/src/ChartPlugin.js
amcgee: @edoardo is it OK to leave this as a comment until @martinkrulltott uses it?  Otherwise we get ESLint errors saying that the prop is defined but never used

packages/plugin/src/modules/getRequestOptions.js
amcgee: Agreed, but we should do that in a separate PR!

packages/plugin/src/ChartPlugin.js
edoardo: Sure!

amcgee:
Thanks @edoardo for the review!  I'm going to also bump @dhis2/analytics here once dhis2/analytics#310 is merged, then merge this one

PR #737, opened 2020-02-26
fix: avoid casting undefined to Number -> NaN

edoardo:
Only cast the value for numeric options if it's defined.
This prevents those options to passed around with NaN as value.
undefined is the default value, and is now passed around as is.


jenniferarnesen:
@edoardo Is there a Jira issue number for this one?

edoardo:
@jenniferarnesen no, just something @janhenrikoverland noticed, we didn't open a ticket for it.

PR #779, opened 2020-03-03
fix: load /#/currentAnalyticalObject crash

edoardo:
Before:
Screenshot 2020-03-03 at 15 06 26

The error above was generated when navigating to `/#/currentAnalyticalObject` which is what happens for example when DV is opened from the Maps app.

The problem turned out to be in `@dhis2/analytics` where after a rename, `defaultVisType` was not exported from the main `index.js`.
It was used in DV, and because of the missing export, it was assuming `undefined` value.
This caused the error above.

The call to `setVisualization` has been removed and  cleaning of visualization and current have been added instead.
This is to avoid that when opening the currentAnalyticalObject after another AO has been loaded,
the store contains a "dirty" state with a visualization object not matching current and ui.
When opening `currentAnalyticalObject`, the app should be in the same state as if a new visualization has been created (not saved) and the Update button has been clicked to render it.


packages/app/src/components/App.js
```diff
@@ -81,7 +81,12 @@ export class App extends Component {
                     getParentGraphMapFromVisualization(AO)
                 )
 
-                this.props.setVisualization(AO)
```
martinkrulltott: The description only mentions that the fix is about updating the Analytics version. @edoardo could you explain why the `setVisualization` was removed? Is it being replaced by `clearVisualization` / `clearCurrent`?

packages/app/src/components/__tests__/App.spec.js
```diff
@@ -137,9 +138,10 @@ describe('App', () => {
                 ).toBeCalledTimes(1)
 
                 expect(props.addParentGraphMap).toBeCalledTimes(1)
-                expect(props.setCurrentFromUi).toBeCalledTimes(1)
```
martinkrulltott: Shouldn't `setCurrentFromUi` still be called? Only `setVisualization` was removed, right?

edoardo: Yes, the test is a couple of lines below.

PR #783, opened 2020-03-05
fix: do not use options intended for download request DHIS2-8401

edoardo:
`hideEmptyRows` and `hideEmptyColumns` options are only intended for the
analytics download request when `tableLayout` is passed (download .xls, .csv and .html+css for PT), not for the normal request used to render the table.

Passing them does not affect the response payload, but does negatively affect caching, both client and server side, because the options are passed in the query string.

Solution is to mark them as non-requestable in the options list config file.

The URLs used in the download menu do not look at this flag, but rely only on the options been set in the visualization object (`current` in the app's Redux store).

Fixes: https://jira.dhis2.org/browse/DHIS2-8401.

Note: the changes in `options.js` for plugin related to `legendDisplayStrategy` and `legend` are there just to keep the file aligned with the one in the app package.
This file should be shared, but for now it's just copied and we should keep them in sync to avoid confusion about which one is the "correct" one.

### DIFF
--- a/packages/plugin/src/modules/options.js
+++ b/packages/plugin/src/modules/options.js
@@ -87,7 +87,7 @@ export const options = {
     subtitle: { defaultValue: undefined, requestable: false, savable: true },
     title: { defaultValue: undefined, requestable: false, savable: true },
 
-    // only for PT XXX
+    // only for PT
     colTotals: { defaultValue: false, requestable: false, savable: true },
     colSubTotals: { defaultValue: false, requestable: false, savable: true },
     rowTotals: { defaultValue: false, requestable: false, savable: true },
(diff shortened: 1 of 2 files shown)


packages/plugin/src/modules/options.js
```diff
@@ -97,14 +97,18 @@ export const options = {
...
+        savable: true,
+    },
+    hideEmptyRows: { defaultValue: false, requestable: false, savable: true },
     skipRounding: { defaultValue: false, requestable: true, savable: true },
     numberType: { defaultValue: 'VALUE', requestable: false, savable: true },
     showHierarchy: { defaultValue: false, requestable: true, savable: true },
     legendSet: { defaultValue: undefined, requestable: false, savable: true },
     legendDisplayStrategy: {
-        defaultValue: undefined,
+        defaultValue: 'FIXED',
```
martinkrulltott: What does `legendDisplayStrategy` and `legend` have to do with this fix?

edoardo: Nothing, but we should keep the 2 files aligned until we can use a shared one, otherwise it will get confusing to know which one is the "right" one.

janhenrikoverland: @martinkrulltott one is in /app, the other is in /plugin

PR #785, opened 2020-03-06
fix: fix loading translations for share/translate dialogs (DHIS2-8249)

edoardo:
Fixes https://jira.dhis2.org/browse/DHIS2-8249.

d2.init must be called after the properties files are added to i18n config, otherwise they are not requested.

Before:
Screenshot 2020-03-06 at 13 57 59
Screenshot 2020-03-06 at 13 58 06

After:
Screenshot 2020-03-06 at 13 58 18
Screenshot 2020-03-06 at 13 58 25


packages/app/src/D2Shim.js
```diff
@@ -11,26 +11,28 @@ import { apiFetchOrganisationUnitLevels } from './api/organisationUnits'
...
+        d2Config.i18n.sources.add(
+            `./i18n_old/i18n_module_${uiLocale}.properties`
+        )
+    }
 
     d2Config.i18n.sources.add('./i18n_old/i18n_module_en.properties')
 }
 
 const initD2 = async ({ baseUrl, apiVersion }, initConfig) => {
+    const userSettings = extractUserSettings(await getUserSettings())
```
amcgee: This won't work because `d2` hasn't been initialized with the correct baseUrl (it gets initialized 2 lines below)

amcgee: Or at least this *shouldn't* work.... were you testing on localhost in development?

edoardo: Yes, I tested on localhost.
I think it can also be that it works for english as that's the fallback anyway.

Tested now after switching to a different language and saw the invalid request to `userSettings` which was not setting the locale, thus no propeties file was loaded for the user locale.

I looked again at the code in v33 and added a commit with a fix.
Now it basically does the same as v33, where `d2.config.baseUrl` is set before calling `userSettings`.

amcgee: Nice, yeah I think that should work!

PR #788, opened 2020-03-09
fix: confirmation before navigating away from unsaved state (DHIS2-8432)

martinkrulltott:
Helps the user avoid losing unsaved work without annoying them by alerting too often. Implements confirmation checks for edited AOs, i.e. previously saved AOs that have new unsaved changes. A.k.a. 'dirty AOs'. https://jira.dhis2.org/browse/DHIS2-8432

For 'dirty AOs':
* Displays a browser `beforeunload` confirmation message when trying to go to a different URL, close the window/tab, go back to another URL etc.
* Displays a custom `ui-core` `Modal` when clicking 'Open' or 'New'

------------

Browser built-in confirmation

image

------------

Custom modal

image

### DIFF
--- a/packages/app/src/components/App.js
+++ b/packages/app/src/components/App.js
@@ -27,6 +27,11 @@ import './scrollbar.css'
 import { getParentGraphMapFromVisualization } from '../modules/ui'
 import AxisSetup from './AxisSetup/AxisSetup'
 import { APPROVAL_LEVEL_OPTION_AUTH } from './VisualizationOptions/Options/ApprovalLevel'
+import {
+    DEFAULT_VISUALIZATION,
+    sGetVisualization,
+} from '../reducers/visualization'
+import { DEFAULT_CURRENT } from '../reducers/current'
 
 export class App extends Component {
     unlisten = null
@@ -114,7 +119,7 @@ export class App extends Component {
 
     componentDidMount = async () => {
         const { store } = this.context
-        const { d2, userSettings } = this.props
+        const { d2, userSettings, visualization, current } = this.props
 
         await store.dispatch(
             fromActions.fromSettings.tAddSettings(userSettings)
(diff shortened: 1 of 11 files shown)


packages/app/src/components/App.js
```diff
@@ -152,6 +174,19 @@ export class App extends Component {
                 e.ctrlKey === true &&
                 this.props.setCurrentFromUi(this.props.ui)
         )
+
+        const t = this
```
janhenrikoverland: Not necessary as you are using an arrow function in the event listener? You should be able to just do this.props.visualization etc inside the arrow function.

packages/app/src/components/App.js
```diff
@@ -210,6 +245,40 @@ export class App extends Component {
                         )}
                     </div>
                 </div>
+                {this.state.locationToConfirm && (
+                    <Modal small>
+                        <ModalTitle>Discard unsaved changes?</ModalTitle>
+                        <ModalContent>
+                            Are you sure you want to leave this visualization?
```
janhenrikoverland: i18n?

packages/app/src/components/App.js
martinkrulltott: I thought the same but it didn't work, I tried multiple combinations but this was the only way to make it work....

janhenrikoverland: Weird, seems to work here

janhenrikoverland: I get t === this // true

packages/app/src/components/App.js
martinkrulltott: Yes, forgot. Added now.

packages/app/src/components/App.js
martinkrulltott: Weird, also working for me now. Before I got `this` but it was different to the `this` I wanted. Maybe there was a cache issue when I tested ⚡️ anyway, updated!

PR #803, opened 2020-03-18
refactor: consolidated various generic error titles to one const (NOJIRA)

martinkrulltott:
Noticed that there was three identical copies of a generic error title "something went wrong", so I consolidated all three to a const that already existed in the error file, to be used by all: https://github.com/dhis2/data-visualizer-app/blob/ea7c1cc9550e43fbe41319ca7672758c7db25f37/packages/app/src/modules/error.js#L247

(NOJIRA) because I just found it ad-hoc and can't really be tested 🤷‍♂ 
No screenshots since its not a visual change 🙂


martinkrulltott:
Opening and closing to test our GitHub - Slack integration

martinkrulltott:
Reopen again 🙂

PR #807, opened 2020-03-18
fix: upgrade analytics for chart title fix [DHIS2-8473]

jenniferarnesen:
Fixed:
Data visualizer supports selection of orgunit levels/groups even if no parent/boundary orgunits are selected. However, the chart title does not seem to support this as it displays "in" both with and without parent orgunits. [DHIS2-8473](https://jira.dhis2.org/browse/DHIS2-8473)

Fixed by upgrading to [Analytics v4.3.24](https://github.com/dhis2/analytics/pull/372)

Example with parent orgunit: "District levels in Sierra Leone"

Without parent orgunit it says: "District levels in"

image

After fix:

image

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -12,7 +12,7 @@
         "redux-mock-store": "^1.5.3"
     },
     "dependencies": {
-        "@dhis2/analytics": "^4.3.23",
+        "@dhis2/analytics": "^4.3.24",
         "@dhis2/app-runtime": "*",
         "@dhis2/d2-i18n": "*",
         "@dhis2/d2-ui-core": "^6.5.5",
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -12,7 +12,7 @@
         "redux-mock-store": "^1.5.3"
     },
     "dependencies": {
-        "@dhis2/analytics": "^4.3.22",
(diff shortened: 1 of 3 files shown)


martinkrulltott:
Sorry I'm a bit late here, since the Analytics change was already approved and merged, but a comment about this: https://github.com/dhis2/analytics/blob/e1a76e05a0fe03b19ac248f703ad8fa2ebcb7721/src/modules/getOuLevelAndGroupText.js#L65
`return ${allDynamicOuNames} ${
            isLevel ? i18n.t('levels') : i18n.t('groups')
        }`

Similar cases have been discussed before where we've come to the conclusion that we should avoid string interpolation / template literals with i18n, as it removes the context for the translators (e.g. `groups` can be translated in different ways depending on the context). Instead we should use the built-in[ i18n interpolation](https://www.i18next.com/translation-function/interpolation). Originally pointed out and suggested by @amcgee I think.

In this specific case it might mean some duplication as we need to display different results based on `isLevel`, which might be bad for the code. But the end result for the translator will be much better, which I think is worth more.

jenniferarnesen:
@martinkrulltott PR is now updated with the latest from analytics, which includes the changes you mentioned above.

PR #809, opened 2020-03-19
refactor: rewrite download menu [DHIS2-8493]

edoardo:
Fixes: https://jira.dhis2.org/browse/DHIS2-8493

Rewrite the Download menu to use `ui-core` in place of material UI.
There are a couple of things still left to replace:

1.  SVG icons are still using material UI icons
2. Download toggle button is still based on material UI. The equivalent in `ui-core` is not available yet, but is in the pipeline, I had a chat with Joe about this and there will be a `<ToolbarButton>` component that is for exactly that purpose (basically a button that does not look like a primary/secondary button). **Update:** this is addressed by https://github.com/dhis2/data-visualizer-app/pull/879 in the meantime.

For point 1 we just have to decide where to put those icons, if in the app or somewhere in `analytics`.
Joe told me that eventually icons will be part of `ui-icons` but until then we have to deal with them ourselves.
For point 2, it can take a while, we just have to wait for team apps.
I suppose we can decide on point 1 and merge this first part and whenever things for point 2 are available, we implement that in a separate PR.

Things left to do/verify:
1. ~A small problem I noticed is that when you open the menu, you can only close it by clicking on the Download button again, clicking outside the menu does not close it.
I'm going to ask @dhis2/team-apps if there is a suggested way to implement Menu close when clicking outside of it.~
This has been solved now by replicating what `DropdownButton` is doing.
Basically using a React portal, `Popper` and a simulation of the `Backdrop` component (not exported ATM).

2. Review styling. The `Menu` and `MenuItems` use the `dense` prop. The only custom styling is for the Menu section titles, the rest is coming straight from `ui-core` (including the slightly awkward looking arrow for the sub-levels). @cooper-joe of course will have the last look when the PR is ready.

Screenshots
Before:
Screenshot 2020-03-19 at 16 30 11

After:
Screenshot 2020-03-19 at 16 29 32


jenniferarnesen:
the menu items look a little squeezed on the right side (too close to the edge of the menu). Here's a screenshot with the padding shown:

image

packages/app/src/components/DownloadMenu/DownloadMenu.js
```diff
@@ -118,222 +89,191 @@ export class DownloadMenu extends Component {
...
-                <ListItemText>{i18n.t('Image (.png)')}</ListItemText>
-            </MenuItem>
-            <MenuItem onClick={this.downloadImage('pdf')}>
-                <ListItemIcon>
-                    <PictureAsPdfIcon />
-                </ListItemIcon>
-                <ListItemText>{i18n.t('PDF (.pdf)')}</ListItemText>
-            </MenuItem>
-        </div>
+        <Fragment>
```
jenniferarnesen: I think you can just use <> now, instead of needing to write <Fragment>.

packages/app/src/components/DownloadMenu/DownloadMenu.js
```diff
@@ -118,222 +89,191 @@ export class DownloadMenu extends Component {
...
+            <div className={styles.menuSectionTitle}>{i18n.t('Graphics')}</div>
+            <MenuItem
```
jenniferarnesen: What about making a local <DenseMenuItem> to remove the repetition of the "dense" property? This is just a matter of taste, not a required change.

jenniferarnesen:
Regarding the icons, maybe we can have them in analytics? I think there are already icons there.

Regarding the mui button, I think it is fine to continue to have that, and replace it when the ui-core button is ready.

I think we should definitely get the "click outside of menu to hide menu" behavior.

edoardo:
> the menu items look a little squeezed on the right side (too close to the edge of the menu). Here's a screenshot with the padding shown. I'm wondering if it would look better if the arrow were inside the content box. But I defer to ui/ux people who are more pixel oriented than me:
> 
> image

Yes, this is what I was referring to in the PR's description point 2 `(including the slightly awkward looking arrow for the sub-levels)`.
This is `ui-core` rendering of the item with a sublevel when using the `dense` option.
See: https://ui-core.dhis2.nu/demo/?path=/story/menu--dense.
@cooper-joe is aware of it and might request some changes to make it look better (perhaps a smaller > icon).

packages/app/src/components/DownloadMenu/DownloadMenu.js
edoardo: Fixed.

packages/app/src/components/DownloadMenu/DownloadMenu.js
edoardo: Fixed.

edoardo:
> Regarding the icons, maybe we can have them in analytics? I think there are already icons there.

Yes, actually some are different from the ones in DV app.
At the moment they are only used internally in analytics, so each one needs to be exported.

> Regarding the mui button, I think it is fine to continue to have that, and replace it when the ui-core button is ready.

Agreed.

> I think we should definitely get the "click outside of menu to hide menu" behavior.

Asked to @dhis2/team-apps and it seems we need to implement that in the app, which I believe means keeping the Menu state in the app's store, connect the `DownloadMenu` component and implement a action/reducer that the app main container can use in a click event handler.

edoardo:
I think we can merge this PR before addressing the icons, which affect the whole app, not only this component.

About the arrow styling, @cooper-joe can say if it's ok like this or we need some custom style to have more right margin/padding.

cooper-joe:
> About the arrow styling, @cooper-joe can say if it's ok like this or we need some custom style to have more right margin/padding.

The arrow styling / lack of padding is intentional on these menu items with type `dense`. I agree it could be adjusted, I'll note it and track the issue in `ui-core`.

edoardo:
So I would say we can merge this as is.
The arrow issue is a minor thing anyway and if it needs to be addressed it should be done in `ui-core`.

packages/app/src/components/DownloadMenu/DownloadMenu.js
```diff
@@ -35,40 +24,37 @@ import {
     apiDownloadTable,
 } from '../../api/analytics'
 import MenuButton from '../MenuButton/MenuButton'
+import MoreHorizontalIcon from '../../assets/MoreHorizontalIcon'
+import styles from './styles/DownloadMenu.module.css'
+
+const DenseMenuItem = ({ Icon, children, ...rest }) => (
```
martinkrulltott, line 30: Separate to another file? Or is it ok to keep more than 1 component per file?

yarn.lock
```diff
@@ -1755,7 +1755,7 @@
   dependencies:
     prop-types "^15"
 
-"@dhis2/ui-core@^4.16.0", "@dhis2/ui-core@^4.6.1", "@dhis2/ui-core@^4.9.1":
```
martinkrulltott, line 1758: This should be kept to prevent previous bugs we've had (yarn insists on breaking these up again but all minor versions should resolve to `"4.16.0"`

packages/app/src/components/DownloadMenu/DownloadMenu.js
edoardo: This is not likely going to be used anywhere else in DV, and it's small, so I guess it's ok to keep it here.

yarn.lock
edoardo: Fixed.

PR #811, opened 2020-03-23
fix: display loading spinner on save (DHIS2-8434)

martinkrulltott:
**Issue:** When saving a visualization in DV, the loading spinner isn't shown so there's no indication that the user should wait before proceeding to interact with the app. https://jira.dhis2.org/browse/DHIS2-8434

**How to recreate:** Go to DV and create a visualization. Click File -> Save and type in any name when prompted. While the app is loading, no indication is shown. The area where it shows "Unsaved visualization" can be used as an indicator to tell that it's still loading (gets replaced by the selected name once it's actually saved).

**Suggested solution:** Set the `pluginLoading` to true when the saving is triggered. This will in turn trigger the spinner to be shown.

----------------

**Please note:** This currently work because
a) The API should always return a positive result (verified by @maikelarabori and @luciano-fiandesio)
b) We always trigger a reload after receiving the response above (which effectively hides the spinner again).

If either of a) or b) isn't true, we could end up with an indefinite spinner 

----------------

Fixed:
loading spinner on save


packages/app/src/components/MenuBar/MenuBar.js
```diff
@@ -77,8 +78,10 @@ const mapStateToProps = state => ({
 const mapDispatchToProps = dispatch => ({
     onRenameVisualization: details =>
         dispatch(fromActions.tDoRenameVisualization(details)),
-    onSaveVisualization: (details, copy) =>
-        dispatch(fromActions.tDoSaveVisualization(details, copy)),
+    onSaveVisualization: (details, copy) => {
+        dispatch(fromLoader.acSetPluginLoading(true))
```
jenniferarnesen: What about moving this to the `tDoSaveVisualization` function? That way the flag is more centrally controlled as both the turning on and the turning off of the flag are in the same place.

Note: 
There seems to be an unused redux prop called `isLoading` (acSetIsLoading). Maybe that is the more appropriate name for this flag, since it is used not only for the plugin loading state, but also the api request? Also, I see that tDoLoadVisualization also sets the acSetPluginLoading flag, but does so after the api request (https://github.com/dhis2/data-visualizer-app/blob/master/packages/app/src/actions/index.js#L56), unlike tDoSaveVisualization, which sets the flag before the request. Maybe these should behave the same? And maybe tDoSaveVisualization should also set the flag to false in the catch. What do you think? I know this grows the scope a tiny bit, but it all kind of hangs together.

martinkrulltott: Do you mean `sGetIsLoading` and `acSetLoading`. I initially thought that this was used to control a loading spinner for the full page, but it turns out that neither of these are actually in use (not that I can find in the code in DV anyway). If that's the case then `acSetPluginLoading` could become `acSetLoading` instead.

martinkrulltott: I agree with moving the loading call to `tDoSaveVisualization`, updated

jenniferarnesen: Yes `sGetIsLoading` was the one I was talking about.

PR #814, opened 2020-03-25
fix: replace MUI with ui-core components - loading spinner (TECH-324)

martinkrulltott:
Replaces the MUI loading spinner with the equivalent component from ui-core. https://jira.dhis2.org/browse/TECH-324
Also removed the duplicate `LoadingMask` file.

----------------

After fix (should look pretty much identical to the current version): 
image


jenniferarnesen:
LGTM

PR #816, opened 2020-03-26
refactor: rewrite visualization type selector (TECH-317)

edoardo:
Fixes: https://jira.dhis2.org/browse/TECH-317

Rewrite the VisualizationTypeSelector and related components to use `ui-core` in place of material UI.
The new solution uses `Popper` and `Card` from `ui-core`, while the vis type "buttons" are custom components.

There are a couple of things still left to replace:

1. ~~SVG icons are still using material UI icons.~~
Solved for now by simply removing MUI `SvgIcon` and use a plain React component that renders an svg node. This icons will be removed and imported from `@dhis2/analytics` instead.

Screenshots
Before:
Screenshot 2020-03-26 at 16 02 59

After:
Screenshot 2020-03-26 at 16 03 18

Not really much difference.
The new version opens all the way to the left of the viewport, but @cooper-joe said we can keep it like this as there will be soon more changes to the whole component as described in this ticket: https://jira.dhis2.org/browse/DHIS2-7855


jenniferarnesen:
A few observations:
1. the cursor:pointer is missing both for the initial menu dropdown, and for the menu items. 
2. The file VisualizationTypeSelector.style.js can be deleted, right?
3. Request for future: would it be possible to set the icon the padding based on viewport height in order to fit the entire menu in the viewport without a vertical scrollbar?

Here is what the menu looks like on my laptop:  
image

edoardo:
About the margins, I implemented Joe's design and I had to make sure that for smaller viewports the scrollbar appears and it's possible to scroll the content of the Card.

We are going to have more vis types, so eventually they are going to be too many to fit without scrolling.

The only way I could make the current icons all fit was to remove the 8px margin all together.
We could also make the Card a little wider to fit 5 icons on 1 row instead of 4.

@cooper-joe any thoughts?

edoardo:
> 1. the cursor:pointer is missing both for the initial menu dropdown, and for the menu items.
> 2. The file VisualizationTypeSelector.style.js can be deleted, right?

> Approved, but go ahead and fix 1 and 2 above before merging.

Both fixed.

PR #857, opened 2020-04-03
fix: fix axes options for line and area vis types

edoardo:
While working on the documentation for DV, I noticed "broken" options in the Axes tab for `Line` and `Area` vis types.

Before:
Screenshot 2020-04-03 at 11 24 12

After:
Screenshot 2020-04-03 at 11 26 38


packages/app/src/modules/options/areaConfig.js
```diff
@@ -55,16 +54,20 @@ export default [
         label: i18n.t('Axes'),
         content: [
             {
-                key: 'axes-section-1',
+                key: 'axes-vertical-axis',
+                label: i18n.t('Vertical (y) axis'),
```
martinkrulltott, line 58: Maybe we can centralise these labels in a `const` somehow so we don't have to repeat them everywhere.

martinkrulltott:
✅  Code and screenshots LGTM

PR #903, opened 2020-04-27
chore: auto merge semver:minor

janhenrikoverland:
Auto merge upgrade PRs that match `semver:minor`.


amcgee:
Just saw this notification - probably should be a `chore` not a `fix` since it shouldn't publish a new version on `npm` @janhenrikoverland

PR #904, opened 2020-04-27
refactor: mui core deps

martinkrulltott:
General cleanup in DV to get rid of the last mui core dependencies.

- MuiThemeProvider
- DefaultAxis
- FatalErrorBoundary
- InterpretationsButton
- IconButton

No functional or style changes, everything looks and works exactly the same.

Note: `DefaultAxis` now has both a `style.js` and a `module.css` dep, as refactoring all in to the css module was out of scope (lots of chainreactions in DefaultLayout + everything in ../../styles/). This should be fixed later in a separate PR.

Known issues:
There's a bug with the IconButton (the three dots on a chip) where the first menu item automatically gets selected on open. This will be fixed once `DimensionMenu` in Analytics is refactored to ui-core components.

---------------

Before: 
image

After:
image

---------------

Before:
image

After:
image

---------------

Before:
image

After:
image


jenniferarnesen:
LG. One small fix - the interpretation button doesn't have the pointer cursor on hover. It used to so I assume that should be the same?

PR #916, opened 2020-04-28
refactor: rewrite Chip and TooltipContent components (TECH-322)

edoardo:
Fixes: https://jira.dhis2.org/browse/TECH-322

Rewrite the Chip and TooltipContent (renamed from Tooltip to solve name clashing issues with ui-core/Tooltip) components to use ui-core in place of material UI.

The ui-core Tooltip available positioning under the element is centered under it, not in the left corner.

A small issue is present when the Chip is dragged, and that is the Tooltip remains on the screen for the close timeout time.
I don't think there is a way to avoid this in the app.
What we might want is that as soon as the drag starts, the Tooltip is removed.
ezgif com-video-to-gif

Screenshots
Before:
Screenshot 2020-04-28 at 15 00 11

After:
Screenshot 2020-04-28 at 14 59 34

### DIFF
--- a/packages/app/src/components/Layout/__tests__/TooltipContent.spec.js
+++ b/packages/app/src/components/Layout/__tests__/TooltipContent.spec.js
@@ -1,31 +1,24 @@
 import React from 'react'
 import { shallow } from 'enzyme'
-import Popper from '@material-ui/core/Popper'
-import { Tooltip } from '../Tooltip'
+import { TooltipContent } from '../TooltipContent'
 
-describe('Tooltip', () => {
+describe('TooltipContent', () => {
     let props
     let shallowTooltip
     const tooltip = () => {
         if (!shallowTooltip) {
-            shallowTooltip = shallow(<Tooltip {...props} />)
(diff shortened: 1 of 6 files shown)


packages/app/src/components/Layout/__tests__/TooltipContent.spec.js
```diff
@@ -1,33 +1,26 @@
...
-            open: true,
-            anchorEl: {},
             dimensionId: 'abc',
             metadata: {},
             itemIds: [],
         }
         shallowTooltip = undefined
     })
 
-    it('renders a Popper', () => {
```
martinkrulltott, line 27: This could be rewritten to "renders a UL" and reused again.

martinkrulltott:
Personally not too bothered by the tooltip-drag-delay bug, but if it persists when merging then please add a JIRA ticket and link it here.

packages/app/src/components/Layout/__tests__/TooltipContent.spec.js
edoardo: Done.

PR #925, opened 2020-04-29
fix: display 'and x others...' on tooltip (DHIS2-8753)

martinkrulltott:
Fixes [8753](https://jira.dhis2.org/browse/DHIS2-8753) (_No tooltip displayed when items exceed window length_), by reducing the max amount of items displayed in a tooltip to 5 and adding a `and X others...` message to replace the subsequent items.

1-5 items display as normal
6 items display` and X other...`
6+ items display `and X others...`

Note: shouldn't affect other tooltips, like the SV one which has its own logic. Same for no items.

----------------

**Examples**

50 items: `and 45 others...`

image

6 items: `and 1 other...`

image

5 items: just displays as normal

image

Single value vis type: unaffected
image

No items: unaffected

image

### DIFF
--- a/packages/app/src/components/Layout/Tooltip.js
+++ b/packages/app/src/components/Layout/Tooltip.js
@@ -101,11 +101,12 @@ export class Tooltip extends React.Component {
                     key={`${this.props.dimensionId}-render-limit`}
                     style={styles.item}
                 >
-                    {`And ${itemDisplayNames.length - renderLimit} ${
-                        itemDisplayNames.length - renderLimit > 1
-                            ? 'others'
-                            : 'other'
-                    }...`}
+                    {itemDisplayNames.length - renderLimit === 1
+                        ? i18n.t('And 1 other...')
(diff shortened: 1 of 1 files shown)


packages/app/src/components/Layout/Tooltip.js
```diff
@@ -81,12 +81,37 @@ export class Tooltip extends React.Component {
...
+                </li>
+            ))
+
+        if (itemDisplayNames.length > renderLimit) {
+            itemsToRender.push(
+                <li
+                    key={`${this.props.dimensionId}-render-limit`}
+                    style={styles.item}
+                >
+                    {`And ${itemDisplayNames.length - renderLimit} ${
```
edoardo, line 104: Translations are missing here, and this won't work properly for all languages.
1) The string should not contain the number, that should be interpolated.
    The problem is that in some languages the position of the number is not the same as for English.
2) plurals are tricky:
    https://www.i18next.com/translation-function/plurals

PR #967, opened 2020-05-11
fix: translations need to be determined at render time [v33] [DHIS2-8638]

jenniferarnesen:
Fixes: https://jira.dhis2.org/browse/DHIS2-8638

Translations need to be figured at render-time, rather than just being consts outside of functions. 3rd party libs (analytics, d2-ui components) also updated with the same type of changes.
Also, datasets.js and datatypes.js were removed as they had previously been migrated to the analytics library (but forgotten to delete here). This resulted in removal of some translations from en.pot (not sure if this means some languages will lose their translations?)


packages/app/src/modules/yearOverYear.js
```diff
@@ -2,36 +2,37 @@ import i18n from '@dhis2/d2-i18n';
...
+        name: () => currentYear - i,
+    }));
 };
 
 // Options for the year on year series dropdown
 export const seriesOptions = [
-    { id: 'THIS_YEAR', name: i18n.t('This year') },
-    { id: 'LAST_YEAR', name: i18n.t('Last year') },
-    { id: 'LAST_5_YEARS', name: i18n.t('Last 5 years') },
+    { id: 'THIS_YEAR', name: () => i18n.t('This year') },
```
martinkrulltott, line 14: I know that this is just a quick fix for v33 and that most of this might have changed in later versions, but shouldn't all of these "refactor prop to function changes" also change the name of the prop? e.g. to **getName** for `name: () => i18n.t('This year')` instead of just **name.** To prevent these functions being disguised as regular props (you'd expect a prop called `name` to be used by just calling `x.name`, not `x.name()`. And `getName` to be called using `x.getName()` 🤷‍♂️ )

packages/plugin/package.json
```diff
@@ -5,7 +5,7 @@
     "main": "./build/index.js",
     "license": "BSD-3-Clause",
     "dependencies": {
-        "@dhis2/analytics": "2.4.9",
+        "@dhis2/analytics": "2.4.12",
```
martinkrulltott, line 8: Should this be part of a new version release of the plugin?

martinkrulltott: or can we start using tilde patch `~` as per [my comment here](https://github.com/dhis2/dashboards-app/pull/822#discussion_r424260469)?

janhenrikoverland: I think adding `~` would be the right thing to do :+1:

martinkrulltott:
I tested this locally, here are my findings. Please note that I haven't verified if there actually are translations available for all of these parts, maybe that should be a first step in the investigation.

----------

image
Caused by the period selector from Analytics?

----------

image

----------

image

-----------
Download menu

image
image


------------
Dimension context menus. Note: the context menu for unselected dimensions seems translated and fine, only selected dimensions have this issue.

image

image

yarn.lock
```diff
@@ -9319,11 +9286,16 @@ minimist@0.0.8:
   resolved "https://registry.yarnpkg.com/minimist/-/minimist-0.0.8.tgz#857fcabfc3397d2625b8228262e86aa7a011b05d"
   integrity sha1-hX/Kv8M5fSYluCKCYuhqp6ARsF0=
 
-minimist@1.2.0, minimist@^1.1.0, minimist@^1.1.1, minimist@^1.1.3, minimist@^1.2.0:
```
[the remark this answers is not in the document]
martinkrulltott: Better than before, but still multiple versions. Good enough though 🙂

yarn.lock
```diff
@@ -137,6 +137,15 @@
     lodash "^4.2.0"
     to-fast-properties "^2.0.0"
 
+"@babel/types@7.8.3":
```
martinkrulltott, line 140: Duplicate dep

PR #968, opened 2020-05-12
fix: transifex config for properties file - sharing dlg

jenniferarnesen:
With this config, we will be able to get translations for the sharing dialog, a d2-ui component that still uses the d2.i18n translation functionality


jenniferarnesen:
Handled in a separate PR

PR #975, opened 2020-05-13
chore: upgrade analytics and use tilde version for deduping in apps [v33]

jenniferarnesen:
Use tilde rather than hat version because v33 needs to stay on 2.4.x of analytics

### DIFF
--- a/packages/plugin/package.json
+++ b/packages/plugin/package.json
@@ -5,7 +5,7 @@
     "main": "./build/index.js",
     "license": "BSD-3-Clause",
     "dependencies": {
-        "@dhis2/analytics": "2.4.9",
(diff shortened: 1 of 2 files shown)


packages/plugin/package.json
```diff
@@ -1,11 +1,11 @@
 {
     "name": "@dhis2/data-visualizer-plugin",
-    "version": "33.1.8",
+    "version": "33.1.10",
     "description": "DHIS2 Data Visualizer plugin",
     "main": "./build/index.js",
     "license": "BSD-3-Clause",
     "dependencies": {
-        "@dhis2/analytics": "2.4.9",
+        "@dhis2/analytics": "^2.4.12",
```
janhenrikoverland: Use tilde

PR #999, opened 2020-05-18
feat: drill down/up for pivot tables

edoardo:
Part of the work for adding drilling feature to PT: https://jira.dhis2.org/browse/DHIS2-8769

This requires PR https://github.com/dhis2/analytics/pull/432 for the click handler on pivot table value cells.

The drilling for org units is simplified here to only have (when available) 1 level up and 1 level down from the org unit that defines the cell value.

Screenshots:
Screenshot 2020-05-20 at 12 04 02

Issues:
- ~Popper does not open in the right position, should open on the right of the clicked cell~
- ~the Popper content should probably use a smaller font @cooper-joe ?~

Things left to do:
- mostly testing and checking the styling (ie. what to do when the MenuItem label is too long, is text wrap ok?)


cooper-joe:
> the Popper content should probably use a smaller font @cooper-joe ?

Yes. Is this a `ui-core` `menu`? If so this should use the `dense` variation.

edoardo:
> Yes. Is this a `ui-core` `menu`? If so this should use the `dense` variation.

Yes, I actually used the `dense` variation in the sublevels but not in the main one.
Fixed now.

packages/plugin/src/VisualizationPlugin.js
```diff
@@ -19,10 +26,27 @@ export const VisualizationPlugin = ({
     onError,
     onLoadingComplete,
     onResponsesReceived,
+    onDrill,
     ...props
 }) => {
     const engine = useDataEngine()
     const [fetchResult, setFetchResult] = useState(null)
+    const [contextualMenuRef, setContextualMenuRef] = useState(undefined)
```
amcgee, line 34: This could be `useRef` instead of `useState`

packages/plugin/src/VisualizationPlugin.js
```diff
@@ -19,10 +26,27 @@ export const VisualizationPlugin = ({
...
+    const [contextualMenuConfig, setContextualMenuConfig] = useState({})
+    const [ouLevels, setOuLevels] = useState(null)
+
+    const onToggleContextualMenu = (ref, data) => {
+        setContextualMenuRef(ref)
+        setContextualMenuConfig(data)
+    }
+
+    const closeContextualMenu = () => setContextualMenuRef(undefined)
```
amcgee, line 43: With `useRef`, `const closeContextualMenu = () => contextualMenuRef.current = undefined`

packages/plugin/src/VisualizationPlugin.js
```diff
@@ -113,34 +155,59 @@ export const VisualizationPlugin = ({
...
-                extraOptions={fetchResult.extraOptions}
-                legendSets={fetchResult.legendSets}
-                {...props}
-            />
-        )
-    }
+    const contextualMenuRect =
+        contextualMenuRef &&
+        contextualMenuRef.current &&
+        contextualMenuRef.current.getBoundingClientRect()
```
amcgee, line 161: With `useRef`:
`const constextualMenuRect = contextualMenuRef.current?.getBoundingClientRect()`

packages/plugin/src/VisualizationPlugin.js
```diff
@@ -19,10 +26,27 @@ export const VisualizationPlugin = ({
...
     ...props
 }) => {
     const engine = useDataEngine()
     const [fetchResult, setFetchResult] = useState(null)
+    const [contextualMenuRef, setContextualMenuRef] = useState(undefined)
+    const [contextualMenuConfig, setContextualMenuConfig] = useState({})
+    const [ouLevels, setOuLevels] = useState(null)
+
+    const onToggleContextualMenu = (ref, data) => {
+        setContextualMenuRef(ref)
```
amcgee, line 39: With `useRef`, `contextualMenuRef.current = ref.current`

packages/plugin/src/VisualizationPlugin.js
```diff
@@ -54,8 +78,26 @@ export const VisualizationPlugin = ({
...
+    useEffect(() => {
+        const doFetch = async () => {
+            const orgUnitLevels = await doFetchOuLevelsData()
+
+            setOuLevels(orgUnitLevels)
+        }
+
+        doFetch().catch(error => onError(error))
+        /* eslint-disable-next-line react-hooks/exhaustive-deps */
+    }, [])
```
amcgee, line 96: This should be `useDataQuery` instead of `dataEngine`.  The line `const [ouLevels, setOuLevels] = useState(null)` and all of this block should be replaced with:

```js
const { data: ouLevelsResponse } = useDataQuery(orgUnitLevelsQuery, { onError })
const ouLevels = ouLevelsResponse?.orgUnitsLevels.organisationUnitLevels
```

There is a slight functional difference here - if the component unmounts before the request completes with the code as written it will try to update the state of an unmounted component, whereas `useDataQuery` will abort the request.

edoardo: Fixed.

packages/plugin/src/VisualizationPlugin.js
edoardo: I tried this approach but it doesn't work for closing the contextual menu when clicking outside of it.
I think the reason is that the component does not re-render when the `contextualMenuRef.current` is reset and the check on `contextualMenuRect` done for showing the `ContextualMenu` component is not performed.
With `useState` instead the component re-renders when the state changes.

amcgee: Hmm... yeah I could see that being a problem.  I think it might be re-rendering any time the cell pivot table re-renders because otherwise I don't think the ref container would change?  But anyway not a big deal, this works as-is so fine for now, wasn't an important suggestion!

amcgee: You could change this to a single line with `?.` optional chaining, but totally optional (😆 )

janhenrikoverland:
Were we able to do something about the menu item width?

Screenshot from 2020-05-25 12-26-20

edoardo:
> Were we able to do something about the menu item width?
> 
> Screenshot from 2020-05-25 12-26-20

Fixed with a workaround.
`ui-core` v5 should automatically expand to maxWidth independently each level.

PR #1009, opened 2020-05-20
fix: upgrade dependencies and fix imports for proper translations [34]

amcgee:
Upgrades cli-app-scripts, analytics, and some d2-ui deps.

Headerbar fixes:
DHIS2-8809 - escape regex special chars in search
DHIS2-8525, DHIS2-8723 - correctly select and use user locale and application title text

There is still VERY patchy coverage of translations in the app itself, and it seems the analytics ones aren’t working, but at least the headerbar and app-local translations seem much improved

PR #1014, opened 2020-05-20
fix: complete translations [v34]

jenniferarnesen:
Part of fix for https://jira.dhis2.org/browse/DHIS2-8894

Please give the app a quick test.

Notes:
*Most of the work was simply switching to use functions to return the result of `i18n.t`.
* The period selector translations aren't yet in the analytics package (since those strings haven't been synced to transifex yet). But note that since some of the period translations are in DV due to  modules/metadata.js, the period dialog appears partially translated.
* yarn.lock was regenerated due to weird errors that seemed to be due to 3rd party deps (no, nuking node_modules didn't fix the problem)
* In plugin, the option for `noData` was removed since analytics now sets the default noData option to be "No data".

Some screenshots:

Note for interpretations that some translations are still missing in the d2-ui-interpretations package. When that is available, then a new PR for DV will be made with the package upgrade.


jenniferarnesen:
@amcgee This PR encompasses all the changes in https://github.com/dhis2/data-visualizer-app/pull/1009.

packages/plugin/src/modules/fetchData.js
```diff
@@ -17,7 +16,6 @@ export const fetchData = async ({
 
     const extraOptions = {
         dashboard: forDashboard,
-        noData: { text: i18n.t('No data') },
```
amcgee, line 20: Is it safe to remove this...?

jenniferarnesen: Yes, I'm setting that as the default in the analytics lib: here: https://github.com/dhis2/analytics/pull/438/files#diff-ed9ecb480b8d52ffa9ceedc0e9988581R39-R44

packages/app/package.json
```diff
@@ -6,19 +6,19 @@
...
         "@dhis2/app-runtime": "*",
         "@dhis2/d2-i18n": "*",
-        "@dhis2/d2-ui-core": "^6.5.11",
-        "@dhis2/d2-ui-file-menu": "^6.5.11",
-        "@dhis2/d2-ui-interpretations": "^6.5.11",
-        "@dhis2/data-visualizer-plugin": "34.3.31",
+        "@dhis2/d2-ui-core": "^7.0.3",
+        "@dhis2/d2-ui-file-menu": "^7.0.3",
+        "@dhis2/d2-ui-interpretations": "^7.0.3",
+        "@dhis2/data-visualizer-plugin": "^34.3.32-alpha.1",
```
martinkrulltott, line 21: why alpha?

jenniferarnesen: The new patch version hasn't been published yet.

PR #1031, opened 2020-05-29
fix: complete translations master (DHIS2-8894)

jenniferarnesen:
Part of fix for https://jira.dhis2.org/browse/DHIS2-8894

Note ***********
The @dhis2/analytics dependency version is an alpha that was created so that others can try out this branch. This PR is dependent on the merge and release of that analytics code (here: https://github.com/dhis2/analytics/pull/462). As soon as that PR is merged/released, this PR will be updated with the correct @dhis2/analytics version


packages/app/src/modules/yearOverYear.js
```diff
@@ -3,35 +3,36 @@ import i18n from '@dhis2/d2-i18n'
 // Fixed years generator
 const getFixedYears = len => {
     let year = new Date().getFullYear()
-    return new Array(len)
-        .fill(null)
-        .map(() => ({ id: String(year), name: String(year--) }))
+    return new Array(len).fill(null).map(() => {
+        const yearString = String(year--)
```
martinkrulltott, line 7: why a separate `const` instead of just `return { id: String(year), getName: () => String(year--) }`?

martinkrulltott: or is this similar to that other issue recently where you needed a `const` to prevent all items from getting the same name (year)?

yarn.lock
```diff
@@ -2,946 +2,1065 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@7.5.5":
```
martinkrulltott, line 5: I assume all of these changes comes from deduping?

packages/app/src/modules/yearOverYear.js
jenniferarnesen: yep

yarn.lock
jenniferarnesen: Yes, I think so. Some changes are also due to the analytics upgrade

packages/app/src/components/DownloadMenu/DownloadMenu.js
```diff
@@ -143,13 +143,16 @@ export const DownloadMenu = ({
...
+                label={i18n.t('ID')}
+                onClick={downloadData(format, 'UID')}
+            />
+            <DenseMenuItem
+                label={i18n.t('Code')}
                 onClick={downloadData(format, 'CODE')}
             />
             <DenseMenuItem
-                label="Name"
+                label={i18n.t('Name')}
```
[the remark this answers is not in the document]
jenniferarnesen: Got 'em. Thanks for catching those.

packages/app/src/modules/metadata.js
```diff
@@ -1,6 +1,6 @@
 import i18n from '@dhis2/d2-i18n'
 
-export const relativePeriods = {
+const getRelativePeriods = () => ({
```
martinkrulltott, line 3: Might be out of scope for this change, but since it has to do with i18n translated strings I thought it was relevant. Why do we keep a copy of all of these strings here when they could/should be fetched from https://github.com/dhis2/analytics/blob/da9b9d1fba9a0c62532293e2501863e61d33ecdf/src/components/PeriodDimension/utils/relativePeriods.js#L14 ?

martinkrulltott: @janhenrikoverland

janhenrikoverland: @martinkrulltott Now that you added them to analytics they can be removed from the app, but the app must be updated, so sounds like a separate task.

martinkrulltott: New ticket: https://jira.dhis2.org/browse/TECH-384 @janhenrikoverland @jenniferarnesen

jenniferarnesen: I'll work on it separately.

PR #1034, opened 2020-05-29
chore: update workflows

janhenrikoverland:
- Removes the lint workflow
- Moves lint step into the test workflow, it's not a separate step for now to avoid checkout/setup duplication, this can be reconsidered


.github/semantic.yml
```diff
@@ -1,4 +1,2 @@
-titleOnly: true
+titleOnly: false
 commitsOnly: false
```
amcgee, line 2: We wanted `titleOnly: true` right?

PR #1068, opened 2020-06-09
feat: upgrade to ui@5 (TECH-389)

martinkrulltott:
Implements [TECH-389](https://jira.dhis2.org/browse/TECH-389)
Upgrades `ui-core` dep to `ui@5`

Depends on https://github.com/dhis2/analytics/pull/476

- [x] **Bug**: `Modals` don't scroll (i.e. `Options` for `PT` which has a lot of content): https://dhis2.slack.com/archives/CBM8LNEQM/p1591781705227100
- [ ] **Bug**: `Selects` in `Modals` render behind the `Modal`
- [x] **Bug**: `LegendSet` and `ApprovalLevel` broken, see inline `FIXME` comments
- [x] **Bug**: `PT` options for `Limit values` doesn't work when selecting from the dropdowns
- [x] **Bug**: Add To layout button menu, Download menu broken
- [x] **Bug**: Drilldown menu broken, can't go past "change org unit" in the menu

Note:
If testing locally before the bug with selects rendering behind modals is fixed, add `.level-2000 { z-index: 5000 !important; }`to `packages/app/src/components/App.css` to see the selects again.


jenniferarnesen:
LG.

martinkrulltott:
> Approved, but I assume merging won't happen until the remaining bug is fixed?

The only remaining bug (`Bug: Selects in Modals render behind the Modal`) is being discussed on Slack: https://dhis2.slack.com/archives/CBM8LNEQM/p1591968893042800. The solution is out of scope of this fix and seems like it will be solved in the next update of `ui`. So this is ready to merge.

PR #1070, opened 2020-06-09
fix: yoy chart mismatch between layout years, and chart years

jenniferarnesen:
This bug is only on master and has not yet been released to play servers. So no Jira issue.

It was caused by decrementing the year after the function that gets the name, but before the function that gets the id.

P.S. A little yarn cleanup seems to have snuck into this PR when I ran a clean install.

Broken:
image (1)

Fixed:

image


janhenrikoverland:
Just noticed something. Look at the ou chip. In the first image it says "Organisation Unit" and "1 valgt". In the second image it says "Organisasjonsenhet" and "1 selected" :man_shrugging:

PR #1138, opened 2020-07-03
fix: avoid unnecessary plugin call (TECH-412)

janhenrikoverland:
Make sure `renderVisualization` is not run twice on mount.


amcgee:
Code LGTM

PR #1143, opened 2020-07-09
feat: column and line combined (DHIS2-9103, DHIS2-6385)

martinkrulltott:
Implements the new Series tab in options to set axes and types, [DHIS2-9103](https://jira.dhis2.org/browse/DHIS2-9103), [DHIS2-6385](https://jira.dhis2.org/browse/DHIS2-6385)
Design spec: [DHIS2-9054](https://jira.dhis2.org/browse/DHIS2-9054)

_**Depends on https://github.com/dhis2/analytics/pull/515**_

**Changes**
- New tab in `Options` called `Series`
- `Column` and `Line` can now switch chart type per each series item
- `Column`, `Line` and `Bar` can now use 4 axes (previously only 2)
- The global vis type is always displayed first in the type options (e.g. `Column` option for `Column` charts)
- Axes always render in the following order: `3 - 1 - [chart] - 2 - 4`
- All other vis types can't switch chart types or use multiple axes, but will still see the Series tab (with an error message)
- Opening the `Series` tab (for `Column`, `Line`, `Bar`) without any items in `Series` will display an error message
- Changing global vis type will reset any custom item types, but retain custom axes (while changing to a type that supports multiple axes)
- `Line` items will always render on top of `Column` items in the chart
- New color sets for axes: [DHIS2-9088](https://jira.dhis2.org/browse/DHIS2-9088)
- The old `Manage Axes` modal (`AxisSetup ` component) and the button to access it (in the Data chip context menu) has been removed.
- `ui.axes` has been replaced by `ui.options.series` in the store, which now keeps track of both `axis` and `type`
- new `barConfig` and `yoyConfig`, as they now need to be separated from `columnConfig`

-----------

**Known issues / bugs** _(todo: unchecked, fixed: checked)_

- [x] Unwanted zebra striping - [Discussion](https://dhis2.slack.com/archives/CBM8LNEQM/p1587381573184700), ticket: [TECH-416](https://jira.dhis2.org/browse/TECH-416)
- [ ] Axis radios aren't centered - [Discussion](https://dhis2.slack.com/archives/CBM8LNEQM/p1594281403011000), ticket: [TECH-415](https://jira.dhis2.org/browse/TECH-415)

-----------

_Settings for Column (and Line)_
image

_Settings for Bar (type options hidden)_
image

_Error message when series has no items_
image

_Error message when another vis type is used_
image

_Column, multi type, two axes_
image

_Column, multi type, four axes_
image

_Column, multi type, single axis_
image

_Line, multi type, multi axes_
image

_Bar, multi axes_
image


packages/app/src/modules/options.js
```diff
@@ -86,6 +86,7 @@ export const options = {
     sortOrder: { defaultValue: '0', requestable: false, savable: true },
     subtitle: { defaultValue: undefined, requestable: false, savable: true },
     title: { defaultValue: undefined, requestable: false, savable: true },
+    series: { defaultValue: [], requestable: false, savable: true },
```
martinkrulltott, line 89: Wasn't sure about how the new `options` props works, does this look correct @edoardo ?

cooper-joe:
Looks good! Two points:

- Please use `dense` type checkboxes instead of the `regular` variety.

- We'll also need to dive in and try to find out how highcharts is selecting colors from the range we provide. Ideally it should pick from the middle, not the darkest shade. Want to do this here or on another PR?

There are also some minor styles I'd like to adjust, mainly the text in the header of the table, but I think those actually belong to the table from `ui`, so I'll take it up there. It does not block merging this.

packages/app/src/modules/options.js
edoardo: `requestable` is for those options that can be passed to the analytics request.
`savable` is for options that should be passed in the AO when saving it.
`defaultValue` should match the one on the backend if present.

martinkrulltott: Thanks @edoardo , so does this look correct then? or should `series` be `requestable`?

martinkrulltott:
Radios now use `dense`, as per @cooper-joe 's comment above
image

packages/app/src/components/VisualizationTypeSelector/VisualizationTypeSelector.js
```diff
@@ -38,6 +38,15 @@ export const VisualizationTypeSelector = (
...
     const handleListItemClick = type => () => {
         setUi(getAdaptedUiByType({ ...ui, type }))
+        setUiSeries(
+            uiSeries.map(item => {
+                const tempItem = { ...item }
+                if (tempItem.type) {
+                    delete tempItem.type
+                }
+                return tempItem
+            })
```
janhenrikoverland, line 48: You can extract this to the `setUiSeries` function itself by accessing `getState` in `mapDispatchToProps` (example below). Then you don't have to pass `uiSeries` as a prop to the component just to send it back.

Example: https://github.com/dhis2/data-visualizer-app/blob/master/packages/app/src/components/UpdateButton/UpdateVisualizationContainer.js#L59

packages/app/src/modules/options/config.js
```diff
@@ -22,14 +22,18 @@ import areaConfig from './areaConfig'
 import pieConfig from './pieConfig'
 import gaugeConfig from './gaugeConfig'
 import singleValueConfig from './singleValueConfig'
+import barConfig from './barConfig'
+import yoyConfig from './yoyConfig'
```
janhenrikoverland, line 26: I like short names, but `yearOverYearConfig` would be more consistent maybe.

packages/app/src/reducers/index.js
```diff
@@ -43,12 +43,9 @@ export {
...
-export const sGetAxisSetupItems = state =>
-    fromUi.sGetAxisSetup(state).map(obj => {
-        const metadata = fromMetadata.sGetMetadata(state)[obj.id]
-
-        return {
-            ...obj,
-            name: metadata ? metadata.name : '',
-        }
-    })
+export const sGetSeriesSetupItems = state =>
```
janhenrikoverland, line 46: Maybe just `SeriesItems`, not `SeriesSetupItems`

packages/app/src/modules/options.js
janhenrikoverland: No, `series` is not a valid analytics api param

packages/app/src/reducers/index.js
martinkrulltott: I used this name to avoid confusion with `series`, which is a prop in `options` fetched by the API (used a lot in this PR). Compared to the `SeriesSetupItems` which instead is based on the `sGetAxisSetup` in the ui reducer ([here](https://github.com/dhis2/data-visualizer-app/blob/feat/DHIS2-9054-series-options/packages/app/src/reducers/ui.js#L307)). But I'm happy to change it if you insist :v:

janhenrikoverland: Ok no, then that's okay :+1:

packages/app/src/components/VisualizationTypeSelector/VisualizationTypeSelector.js
martinkrulltott: Good point, fixed

packages/app/src/components/VisualizationTypeSelector/VisualizationTypeSelector.js
```diff
@@ -129,6 +131,7 @@ const mapStateToProps = state => ({
 
 const mapDispatchToProps = dispatch => ({
     setUi: ui => dispatch(acSetUi(ui)),
+    clearSeriesTypes: () => dispatch(acClearSeriesType()),
```
janhenrikoverland, line 134: More general name like `onItemClick`, can be fixed later as discussed.

martinkrulltott: Fixed

packages/app/src/components/VisualizationOptions/Options/SeriesTable.js
```diff
@@ -0,0 +1,237 @@
...
+
+const SeriesTable = ({
+    layoutItems,
+    optionItems,
+    onChange,
+    visType,
+    showAxisOptions,
+    showTypeOptions,
+}) => {
+    const availableAxes = [0, 1, 2, 3]
```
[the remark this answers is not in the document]
martinkrulltott: Fixed

PR #1161, opened 2020-07-28
fix: Decimals and Steps options only allow positive values (DHIS2-9002, DHIS2-9194)

martinkrulltott:
Fixes [DHIS2-9002](https://jira.dhis2.org/browse/DHIS2-9002), [DHIS2-9194](https://jira.dhis2.org/browse/DHIS2-9194)

Based on playground [PoC]( https://codesandbox.io/s/ui-positive-numbers-nzhv6?file=/src/App.js)

`Decimals` and `Steps` could previously be entered as negative numbers and decimal values, which was incorrect. This change prevents this behaviour and only allows positive whole numbers, e.g. `0`, `1`, `2`. Negative values will `undefined`, which in turn displays the placeholder. Note that for `Decimals`, `0` and `undefined` outputs the same result (no decimals shown), while for `Steps`, `0` means no steps while `undefined` displays the default amount of steps set by Highcharts.

- New base type `PositiveNumberBaseType` added that only allows positive whole numbers
- `Math.round` prevents decimal numbers and returns `Number` instead of `String,` which fixes [DHIS2-9194](https://jira.dhis2.org/browse/DHIS2-9194)
- `parseInt(value, 10)` to allow for clearing the value with backspace
- Applies to both typing manually and by using the browser's built-in "stepper arrows". 
- New help text for Steps, as per [Slack discussion](https://dhis2.slack.com/archives/G3TL0J145/p1595935310011500?thread_ts=1595923304.005000&cid=G3TL0J145), _Number of axis tick steps, including the min and max. A value of 2 or lower will be ignored._

----------

Example input and expected output:
- Typing whole numbers 0, 1, 2 `->` kept as is
- Typing negative numbers -1 `->` 0
- Decimal numbers 1.6, 1.1 `->` rounded (to 2 and 1 respectively)
- Stepping below 0 `->` 0
- Clearing the input with backspace `->` undefined

----------

ezgif com-video-to-gif


packages/app/src/components/VisualizationOptions/Options/RangeAxisSteps.js
```diff
@@ -2,11 +2,10 @@ import React from 'react'
...
-import TextBaseOption from './TextBaseOption'
+import PositiveNumberBaseType from './PositiveNumberBaseType'
 
 export const RangeAxisSteps = () => (
-    <TextBaseOption
-        type="number"
+    <PositiveNumberBaseType
         width="96px"
         helpText={i18n.t(
             'The number of axis steps between the min and max values'
```
martinkrulltott, line 11: Should this help text be changed? Discussion: https://dhis2.slack.com/archives/G3TL0J145/p1595923304005000

packages/app/src/components/VisualizationOptions/Options/PositiveNumberBaseType.js
```diff
@@ -0,0 +1,64 @@
...
+        <div>
+            <InputField
+                type="number"
+                label={label}
+                onChange={({ value }) => {
+                    const parsedValue = parseInt(value, 10)
+                    parsedValue >= 0
+                        ? onChange(Math.round(value))
+                        : onChange(parsedValue ? 0 : null)
+                }}
```
janhenrikoverland, line 29: There's a small bug here that could make it return negative numbers if the input is between 0 and -1.

E.g. if you pass "-0.7" then `parsedValue >= 0` will be true, but Math.round will return -1.

If you pass "-0.3" so that the value is floored Math.round will return -0 (which is the same as 0, but not so nice).

janhenrikoverland: Suggestion:
`
const parsedValue = parseInt(value, 10)
onChange(isNaN(parsedValue) ? 0 : Math.max(parsedValue, 0))
`

martinkrulltott: Thanks for pointing this out. However when testing (OSX Chrome) the example values above it seems to work as expected and always return `0`. I believe `parsedValue >= 0` prevents anything with `-` and thus reverts to `0` instead. Or am I missing something?

PR #1164, opened 2020-07-29
feat: support for two category charts [DHIS2-7876]

edoardo:
Fixes https://jira.dhis2.org/browse/DHIS2-7876.

Most of the inplementation is in the analytics dependency, see: https://github.com/dhis2/analytics/pull/508.

The DV app needs to correctly configure the `ui.layout`object when loading an AO saved with 2 categories, otherwise the default layout adapter moves all dimensions besides the first from `rows` to `filters`.

### DIFF
--- a/packages/app/src/modules/ui.js
+++ b/packages/app/src/modules/ui.js
@@ -11,6 +11,7 @@ import {
     VIS_TYPE_PIVOT_TABLE,
     defaultVisType,
     isYearOverYear,
+    isDualCategoryChartType,
 } from '@dhis2/analytics'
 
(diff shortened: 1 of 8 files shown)


packages/app/src/modules/ui.js
```diff
@@ -76,6 +84,10 @@ export const singleValueUiAdapter = ui => ({
 })
 
 export const getAdaptedUiByType = ui => {
+    if (isDualCategoryChartType(ui.type) && ui.layout.rows.length > 1) {
```
janhenrikoverland, line 87: Rules vs >1

PR #1171, opened 2020-07-31
fix: disable lines and vertical axis options for multi axes (DHIS2-9010, DHIS2-9013, DHIS2-9014)

martinkrulltott:
Fixes [DHIS2-9010](https://jira.dhis2.org/browse/DHIS2-9010), [DHIS2-9013](https://jira.dhis2.org/browse/DHIS2-9013), [DHIS2-9014](https://jira.dhis2.org/browse/DHIS2-9014)

**Depends on https://github.com/dhis2/analytics/pull/529**

Line and vertical axis options needs to be disabled when using multiple axes (see spec in ticket).

- Disables Lines ([DHIS2-9010](https://jira.dhis2.org/browse/DHIS2-9010)) and Vertical Axis options (title: [DHIS2-9013](https://jira.dhis2.org/browse/DHIS2-9013), options: [DHIS2-9014](https://jira.dhis2.org/browse/DHIS2-9014)) by checking the new Analytics prop `hasCustomAxes`
- Lines and Vertical Axis config sections extracted to own modules for reduced duplication
- Adds a `disabled` prop to various options
- Wraps configs in functions and adds prop `hasCustomAxes` where applicable
- Refactors `getLabel()` to `label`, since config is now wrapped in a function (no need for individual fn wrapping)
- Additionally, adds the correct margin between some options that were missing it
- _B-b-bonus round!_ Changes incorrectly named selector `getAxisIdByDimensionId` to `sGetAxisIdByDimensionId`

--------------------

**TODO**
- [x] Implement the disable logic to all custom axes types (column, line, bar)
- [x] Change all configs to functions, e.g. `const config = hasCustomAxes => [`
    - [x] Change config `getLabel` to `label`?
- [x] Prevent Trend line from rendering in a chart when disabled
- [x] Change the disabled help text to the [help text style](https://ui.dhis2.nu/demo/?path=/story/help--default)
- [x] Invalid type prop console errors
- [x] Replace `alpha` with final Analytics version once released

--------------------

Line options disabled when using multiple axes

image

Vertical axis options disabled when using multiple axes

image

Correct margins for Column Data

image

Correct margins Column Axes

image

Correct margins for Pivot table Legend

image

Correct margins for Pivot Table Style

image

### DIFF
--- a/packages/app/src/components/VisualizationOptions/Options/SelectBaseOption.js
+++ b/packages/app/src/components/VisualizationOptions/Options/SelectBaseOption.js
@@ -22,10 +22,10 @@ export const SelectBaseOption = ({
     disabled,
 }) => {
     const defaultValue = option.defaultValue
-    const [checked, setchecked] = useState(value !== defaultValue)
+    const [checked, setChecked] = useState(value !== defaultValue)
     const selected = option.items.find(item => item.value === value)?.value
     const onToggle = checked => {
-        setchecked(checked)
+        setChecked(checked)
 
         onChange(checked ? option.items[0].value : defaultValue)
     }
--- a/packages/app/src/components/VisualizationOptions/Options/SelectBaseOption.js
+++ b/packages/app/src/components/VisualizationOptions/Options/SelectBaseOption.js
@@ -19,12 +19,13 @@ export const SelectBaseOption = ({
     toggleable,
     value,
     onChange,
+    disabled,
 }) => {
     const defaultValue = option.defaultValue
-    const [enabled, setEnabled] = useState(value !== defaultValue)
+    const [checked, setchecked] = useState(value !== defaultValue)
     const selected = option.items.find(item => item.value === value)?.value
     const onToggle = checked => {
-        setEnabled(checked)
+        setchecked(checked)
 
         onChange(checked ? option.items[0].value : defaultValue)
     }
@@ -33,14 +34,15 @@ export const SelectBaseOption = ({
         <div className={tabSectionOption.className}>
             {toggleable ? (
                 <Checkbox
-                    checked={enabled}
+                    checked={checked}
                     label={label}
                     name={`${option.name}-toggle`}
                     onChange={({ checked }) => onToggle(checked)}
                     dense
+                    disabled={disabled}
                 />
             ) : null}
-            {!toggleable || enabled ? (
+            {(!toggleable || checked) && !disabled ? (
                 <div
                     className={
                         toggleable ? tabSectionOptionToggleable.className : ''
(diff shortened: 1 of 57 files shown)


packages/app/src/modules/options/columnConfig.js
```diff
@@ -5,49 +5,36 @@ import i18n from '@dhis2/d2-i18n'
...
+export default hasCustomAxes => [
     {
         key: 'data-tab',
-        getLabel: () => i18n.t('Data'),
+        label: i18n.t('Data'),
         content: [
             {
                 key: 'data-display',
-                getLabel: () => i18n.t('Display'),
+                label: i18n.t('Display'), //TODO: change to "label" now that the whole config is wrapped in a fn?
```
edoardo, line 27: remove comment?

packages/app/src/modules/options/config.js
```diff
@@ -25,33 +25,33 @@ import singleValueConfig from './singleValueConfig'
 import barConfig from './barConfig'
 import yearOverYearConfig from './yearOverYearConfig'
 
-export const getOptionsByType = type => {
+export const getOptionsByType = (type, hasCustomAxes) => {
     switch (type) {
         case VIS_TYPE_COLUMN:
-            return columnConfig
+            return columnConfig(hasCustomAxes)
```
edoardo, line 31: This potentially needs to be expanded into an object with various flags...

packages/app/src/modules/options/columnConfig.js
martinkrulltott: Thanks, removed!

packages/app/src/components/VisualizationOptions/Options/SelectBaseOption.js
```diff
@@ -19,30 +19,30 @@ export const SelectBaseOption = ({
     toggleable,
     value,
     onChange,
+    disabled,
 }) => {
     const defaultValue = option.defaultValue
-    const [enabled, setEnabled] = useState(value !== defaultValue)
+    const [checked, setchecked] = useState(value !== defaultValue)
```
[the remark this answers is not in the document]
martinkrulltott: Thanks, fixed!

packages/app/src/modules/options/config.js
martinkrulltott: Yes indeed, if more props are needed then this should be changed to an object instead. Hopefully `hasCustomAxes` (which is the same as "has multiple axes") will suffice for the near future though. I think you can use it for hiding the color sets as well, right?

edoardo: Yes, for that it will work.
I'm thinking about the "Legends applied to bar and column charts only one data item should be allowed in the series with only one axis" type of condition 😆

PR #1193, opened 2020-08-14
fix: prevent series from saving global vis type as custom type (DHIS2-6385)

martinkrulltott:
When using multi type charts, a minimalistic approach should be used for saving the items in the `series` array.
If the user selects a custom type, this should only be saved if it differs from the global vis type, as the global vis type will be used per default if no type us present.
Before this fix, if a user clicked a custom type (e.g. line) and then clicked the default value again (e.g. column), the default value would be saved in the store.

Technically, instead of saving the default type as a prop, the whole prop is removed when it matches the global vis type.

-----------

### Examples

**E1**

All values are default (vis type), no `type` prop is saved in the store
image
image

**E2**

One custom type has been set and is the only `type` prop saved in the store
image
image


packages/app/src/components/VisualizationOptions/Options/SeriesTable.js
```diff
@@ -62,11 +64,19 @@ const SeriesTable = ({
...
-        item[prop] = value
+
+        if (prop === TYPE_PROP && value === visType) {
+            const { [prop]: remove, ...rest } = series[itemIndex]
+            series[itemIndex] = { ...rest }
+        } else {
+            series[itemIndex][prop] = value
+        }
+
         onChange(series)
```
janhenrikoverland, line 80: @martinkrulltott To me this kind of work is typically the reducer's responsibility, so to avoid doing all this logic at (a dumb?) component level, including deep cloning which is a bit nasty, maybe create a new action for updating a series object?

janhenrikoverland: (Instead of dispatching the full new series array)

martinkrulltott: Good points, fixed, please see my latest commit 👍

janhenrikoverland (approved):
LG

PR #1210, opened 2020-08-21
fix: Decimals and Steps options only allow positive values (DHIS2-9002, DHIS2-9194) (#1161) v34

martinkrulltott:
Backport of https://github.com/dhis2/data-visualizer-app/pull/1161 for v34

Had to change `ui` to `ui-core`, since v34 is using an older version of this library.

* New base type PositiveNumberBaseType added that only allows positive whole numbers
* Math.round prevents decimal numbers and returns Number instead of String, which fixes DHIS2-9194
* parseInt(value, 10) to allow for clearing the value with backspace
* Applies to both typing manually and by using the browser's built-in "stepper arrows".
* New help text for Steps, as per Slack discussion, Number of axis tick steps, including the min and max. A value of 2 or lower will be ignored.

### DIFF
--- a/packages/app/i18n/en.pot
+++ b/packages/app/i18n/en.pot
@@ -464,7 +464,9 @@ msgstr ""
 msgid "Auto"
 msgstr ""
 
-msgid "The number of axis steps between the min and max values"
+msgid ""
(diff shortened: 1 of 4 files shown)


edoardo (approved):
LGTM

PR #1211, opened 2020-08-21
fix: reorder options for consistency (DHIS2-9004) (519b0ba) v34

martinkrulltott:
Backport https://github.com/dhis2/data-visualizer-app/pull/1153 for v34

Regenerated the pot file in a separate commit.

* Removes two incorrect fields in Style tab and adds missing section titles (areaConfig, lineConfig), as per DHIS2-9004
* Moves ShowData from Data to Style (areaConfig, lineConfig, stackedColumnConfig)
* Adds missing section splitting in the Data tab (areaConfig, lineConfig)
* Changes the label const to instead use the getLabel function (areaConfig, lineConfig)


edoardo (approved):
LGTM

PR #1212, opened 2020-08-21
fix: pass y-axis steps as number (DHIS2-9194) v33

martinkrulltott:
Manual backport of [DHIS2-9194](https://jira.dhis2.org/browse/DHIS2-9194) to v33
The 34 and 35 solution couldn't be cherry-picked as it contained changes not compatible for 33.
However the fix for this issue is simply passing the steps value as a number instead of a string. As mentioned by @janhenrikoverland  here: [comment](https://jira.dhis2.org/browse/DHIS2-9194?focusedCommentId=34665&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-34665)

PR #1216, opened 2020-08-21
fix: pass y-axis steps as number (DHIS2-9194) v32

martinkrulltott:
Backports https://github.com/dhis2/data-visualizer-app/pull/1212 to v32

* fix: pass number value as number


edoardo (approved):
LGTM

PR #1229, opened 2020-08-28
feat: move dx to filter for YOY which allows for multiple dx items (DHIS2-8808)

janhenrikoverland:
Depends on https://github.com/dhis2/analytics/pull/574

Fixes [DHIS2-8808](https://jira.dhis2.org/browse/DHIS2-8808)

The `dx` dimension is now always `filter`, both in the app and the backend. The restriction where only the first item was picked has been removed.

Screenshot from 2020-08-28 10-30-45

### DIFF
--- a/packages/app/src/modules/current.js
+++ b/packages/app/src/modules/current.js
@@ -140,20 +140,21 @@ export const getPieCurrentFromUi = (state, action) => {
 export const getYearOverYearCurrentFromUi = (state, action) => {
     const ui = action.value
 
-    const dxItem = ui.itemsByDimension[DIMENSION_ID_DATA]
-        ? [ui.itemsByDimension[DIMENSION_ID_DATA][0]]
-        : []
+    const periodDimension = dimensionCreate(
+        DIMENSION_ID_PERIOD,
+        ui.yearOverYearCategory
+    )
+
+    const filtersWithoutPeriod = getAxesFromUi(ui).filters.filter(
+        f => f.dimension !== DIMENSION_ID_PERIOD
+    )
 
     return {
         ...state,
         [BASE_FIELD_TYPE]: ui.type,
-        [AXIS_ID_COLUMNS]: [dimensionCreate(DIMENSION_ID_DATA, dxItem)],
-        [AXIS_ID_ROWS]: [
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -1302,13 +1302,11 @@
     debug "^3.1.0"
     lodash.once "^4.1.1"
 
-"@dhis2/analytics@^8.9.6":
+"@dhis2/analytics@^8.9.6", "@dhis2/analytics@file:../analytics":
```
martinkrulltott, line 1305: 🙂

packages/app/src/modules/current.js
```diff
@@ -140,20 +140,21 @@ export const getPieCurrentFromUi = (state, action) => {
...
-    const dxItem = ui.itemsByDimension[DIMENSION_ID_DATA]
-        ? [ui.itemsByDimension[DIMENSION_ID_DATA][0]]
-        : []
+    const periodDimension = dimensionCreate(
+        DIMENSION_ID_PERIOD,
+        ui.yearOverYearCategory
+    )
+
+    const filtersWithoutPeriod = getAxesFromUi(ui).filters.filter(
+        f => f.dimension !== DIMENSION_ID_PERIOD
```
martinkrulltott, line 149: rename `f` to `filter` or `filterItem`?

martinkrulltott: I understand why `DIMENSION_ID_PERIOD` needs to be excluded from the `filters`, but why does `DIMENSION_ID_PERIOD` end up in `getAxesFromUi(ui).filters` to begin with? I thought the period dimension wasn't allowed on YoY in the first place?
Maybe I'm missing something, but just curious if this exclusion is unnecessary? (if period is never present anyway?)

martinkrulltott (approved):
LGTM. Happy to merge as is (once Analytics is released/added). One side-note question though. But since I'm not 100% familiar with `ui` I might just be missing something

packages/app/src/modules/current.js
janhenrikoverland: It can be removed I think, that filter was there already, just to be on the safe side probably.

yarn.lock
janhenrikoverland: Updated

martinkrulltott (approved):

PR #1230, opened 2020-08-28
feat: allow YOY without dx (DHIS2-8806)

janhenrikoverland:
Depends on https://github.com/dhis2/analytics/pull/574 for the feature to work, but can safely be merged without it.

Removes the `dx` requirement for YOY types. If no DEGS or `dx` the server returns a `E7102` which is already handled by DV.


martinkrulltott (approved):

martinkrulltott:
LGTM. However I'd advice to test this locally once https://github.com/dhis2/data-visualizer-app/pull/1229 is merged to make sure everything works as expected before merging this.

PR #1232, opened 2020-08-28
fix: set a default name if saved vis is untitled (DHIS2-9360)

martinkrulltott:
Fixes [DHIS2-9360](https://jira.dhis2.org/browse/DHIS2-9360)

If no name is provided in the Save dialog the save process will now default to a date-based name in the format of `Untitled {{visualizationType}} visualization, {{date}}`.

The date in turn is based on a localized version of `dd mmm yyyy`.

image

### DIFF
--- a/packages/app/src/actions/index.js
+++ b/packages/app/src/actions/index.js
@@ -31,6 +31,7 @@ import {
     GenericServerError,
     VisualizationNotFoundError,
 } from '../modules/error'
+import { getDisplayNameByVisType } from '@dhis2/analytics'
 
 export {
     fromVisualization,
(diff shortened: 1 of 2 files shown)


packages/app/src/actions/index.js
```diff
@@ -187,9 +188,22 @@ export const tDoSaveVisualization = ({ name, description }, copy) => async (
...
+                        visualization.type
+                    ),
+                    date: new Date().toLocaleDateString(undefined, {
+                        year: 'numeric',
+                        month: 'short',
+                        day: '2-digit',
+                    }),
+                }
+            )
         }
```
janhenrikoverland, line 207: Maybe just do `visualization.name = name || .....`, the if/else is not necessary when `name` is being set in both cases.

martinkrulltott: Fair enough, I was just following the if-pattern of the rest of the file. Updated 👍

janhenrikoverland (approved):

PR #1244, opened 2020-09-03
fix: DGS default from settings (DHIS2-9328)

edoardo:
Related to [DHIS2-9328](https://jira.dhis2.org/browse/DHIS2-9328).

This fixes the issue of the DGS default from system settings not being used when resetting the ui portion of the DV's Redux store, which ultimately caused the visualization (SV, PT) to not render the digit group separator correctly when working on a non-saved visualization.

The DGS in the store's ui was only set when the option in the options dialog was rendered/interacted with, so for a new visualization, the DGS was undefined even if it was set to a specific value in system settings.

### DIFF
--- a/packages/app/src/components/VisualizationOptions/Options/DigitGroupSeparator.js
+++ b/packages/app/src/components/VisualizationOptions/Options/DigitGroupSeparator.js
@@ -1,19 +1,14 @@
 import React from 'react'
-import PropTypes from 'prop-types'
-import { connect } from 'react-redux'
 
 import i18n from '@dhis2/d2-i18n'
 
 import SelectBaseOption from './SelectBaseOption'
 
-import { sGetSettingsDigitGroupSeparator } from '../../../reducers/settings'
-
-const DigitGroupSeparator = ({ defaultValue }) => (
+const DigitGroupSeparator = () => (
     <SelectBaseOption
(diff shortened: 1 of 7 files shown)


packages/app/src/components/VisualizationOptions/Options/DigitGroupSeparator.js
```diff
@@ -1,19 +1,14 @@
...
 
-import { sGetSettingsDigitGroupSeparator } from '../../../reducers/settings'
-
-const DigitGroupSeparator = ({ defaultValue }) => (
+const DigitGroupSeparator = () => (
     <SelectBaseOption
         label={i18n.t('Digit group separator')}
         option={{
             name: 'digitGroupSeparator',
-            defaultValue,
```
martinkrulltott, line 16: Will the system setting still be used as the default selected value in the ui?

edoardo: Yes, because option components are connected to the store and read their value from it.
Default values are used to set the `value` prop on the components.
Now this particular option implementation is back to what it was initially, with not default and the value is just read from the store.
So by setting the default value properly in the store whenever the ui portion of it is reset, we can assure we always have the correct default matching the system settings one.

PR #1254, opened 2020-09-09
fix: Decimals and Steps options only allow positive values (DHIS2-9002, DHIS2-9194) (#1161) v33

martinkrulltott:
Backport of https://github.com/dhis2/data-visualizer-app/pull/1161 for v33

Based on cherry-picking https://github.com/dhis2/data-visualizer-app/commit/b0996072769123a02249c63cb41cf347d90ad2f2

Not a clean CP, had to change quite a lot to adapt it to v33, since it uses MUI instead of ui/ui-core.

---------

image

image

---------

Original description:

* New base type PositiveNumberBaseType added that only allows positive whole numbers
* Math.round prevents decimal numbers and returns Number instead of String, which fixes DHIS2-9194
* parseInt(value, 10) to allow for clearing the value with backspace
* Applies to both typing manually and by using the browser's built-in "stepper arrows".
* New help text for Steps, as per Slack discussion, Number of axis tick steps, including the min and max. A value of 2 or lower will be ignored.

### DIFF
--- a/packages/app/i18n/en.pot
+++ b/packages/app/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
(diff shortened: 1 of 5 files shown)


janhenrikoverland (approved):
LG

PR #1256, opened 2020-09-09
fix: rename and invert the hide legend option (#1152) (DHIS2-9003) v33

martinkrulltott:
Backport of https://github.com/dhis2/data-visualizer-app/pull/1152 to 33.x

Based on cherry-picking https://github.com/dhis2/data-visualizer-app/commit/026ae9e77eafebac409b4a5ac3f36b44c333c820

Not a clean CP, had to adapt it to v33, since it uses MUI instead of ui/ui-core.

---------

image

---------

Original description:
* Changes the title of the hideLegend option to a positive case (show).
* As the backend option is a negative case (hide), CheckboxBaseOption now has an inverted prop to flip the checked state.


janhenrikoverland (approved):
LG

PR #1270, opened 2020-09-15
fix: confirm dialog with interpretation view

edoardo:
The changes here are for improving when the navigation confirm dialog is shown and what happens when clicking Yes or No.

Since there is actually no change loss when navigating in and out of an interpretation's view, the confirm dialog does not show, even if there are changes to the visualization.

ezgif com-optimize (2)

When answering No to the confirm dialog, the previous URL is loaded so that the state of the app is really what it was before.

ezgif com-optimize (3)


edoardo:
Integrated into #1273

PR #1271, opened 2020-09-15
fix: allow axis range to be reset (not default to 0)

martinkrulltott:
Issue: When the axis range (min and max) are used and subsequently erased / reset, they default to `0` instead of `undefined`, causing HighCharts to render a weird looking chart.

Solution: Prevent `''` to be cast to `0` by adding a check in `current.js`

----------

_Bug causing both min and max axis range to default to 0 when it's empty_
image

----------

_With the fix the axis range defaults to `undefined` instead, causing the correct chart to be rendered_
image


edoardo (approved):

martinkrulltott:
Moved to 1266

PR #1272, opened 2020-09-15
fix: set default layout to 50-50 width

janhenrikoverland:
Now that we support two-category charts it makes sense to allocate more space to categories in the layout.

Before:
Screenshot from 2020-09-15 14-40-38

This PR fixes this through the smallest possible fix. After:

Screenshot from 2020-09-15 14-40-49

I'm simply changing the default layout. The result is that all chart types that support 2C get this change. The downside is that Radar also gets it, which does not support it, but that is an okay trade-off vs making a separate layout for radar charts at this point. Also, the specific layout for pivot tables could be removed later as it is now identical to the default layout.


martinkrulltott (approved):

janhenrikoverland:
Moved to https://github.com/dhis2/data-visualizer-app/pull/1266

PR #1259, opened 2020-09-09
fix: set a default name if saved vis is untitled (DHIS2-9360) v33

martinkrulltott:
Backports https://github.com/dhis2/data-visualizer-app/pull/1232 to v33

Almost clean cherry-pick of https://github.com/dhis2/data-visualizer-app/commit/f9fc26b9b25dbf9f43a71d13cc67c257db213b1c. Had to change Analytics `getDisplayNameByVisType` 
 to DV's own `getChartTypeDisplayName` fn, as this hadn't been moved to Analytics yet in v33.

-----
Proof
image

-----

In addition, https://github.com/dhis2/data-visualizer-app/commit/41c264e74cb28812c071016fd412015cab8b6935 was cherry-picked to fix the issue with the name getting overwritten when saving a second time.

PR #1273, opened 2020-09-15
fix: keep the visualization name when re-saving after edit

edoardo:
When clicking Save on a stored visualization, keep the same name and
avoid the default Untitled name to be used instead.

Demo:
ezgif com-optimize (4)

---------

Related to saving is also the confirm dialog when navigating in and out of an interpretation's view, which had some issues.

The changes here are for improving when the navigation confirm dialog is shown and what happens when clicking Yes or No.

Since there is actually no change loss when navigating in and out of an interpretation's view, the confirm dialog does not show, even if there are changes to the visualization.

ezgif com-optimize (2)

When answering No to the confirm dialog, the previous URL is loaded so that the state of the app is really what it was before.

ezgif com-optimize (3)


martinkrulltott (approved):

martinkrulltott:
Needs to be backported to v34 (original v34 backport https://github.com/dhis2/data-visualizer-app/pull/1258) and v33 (original v33 backport https://github.com/dhis2/data-visualizer-app/pull/1259/) as well.

PR #1280, opened 2020-09-17
fix: make series options unavailable for relative items

martinkrulltott:
**Depends on https://github.com/dhis2/analytics/pull/605**

To prevent series options from being used with relative items (relative periods, relative org units, assigned categories), series options needs to be disabled for these items.

Relative items will result in the following error message: 

image

Relative items are:

- Assigned Categories
- Periods from the Relative periods tab
- User org units
- Org unit levels and groups

Blocking any of these from presenting the series options is quite straight forward. However, there's a much more complex (edge) case where the user: 

1. add fixed org units or periods (doesn't apply to Assigned Categories)
2. change type and axes in the series options
3. add a relative org unit or period (leaving the initial selection as is)
4. display custom series options for the initial selection and default series options for the relative item(s).

To prevent this case and make sure series options are **always** disabled when any relative items are in use, https://github.com/dhis2/analytics/pull/605 exports the fn `hasRelativeItems` which is used to switch between the regular view and the custom series options view.

DV is responsible for:
1. display an error message for the Series tab when relative items are used
2. re-enable lines and vertical axis options when relative items are used (even though series options contain multiple axes)

Analytics is responsible for:
1. the `hasRelativeItems` logic
2. generating a chart _with single axis_ props when relative items are used (when multi axes would otherwise be used)

-----------

**Example**

-----------

_Settings before a relative item has been added_
image

image

image

_Result_
image

-----------

_Subsequently adding a relative period_
image

_Settings_
image

image

image

_Result_
image

-----------

Fixes: https://jira.dhis2.org/browse/DHIS2-9486.

### Display legend option
Radio and font size in the options' Legend tab didn't use the `dense` variant.

Before:
Screenshot 2020-09-11 at 10 31 36

After:
Screenshot 2020-09-11 at 10 31 52

### Limit values tab
Additional style fix for `Limit values` tab in PT, after checking with @cooper-joe:

Before:
Screenshot 2020-09-16 at 14 50 41

After:
Screenshot 2020-09-16 at 14 51 18


edoardo (approved):

janhenrikoverland (approved):
LG

PR #1298, opened 2020-09-22
fix: legend type spacing in SV options

edoardo:
There was too little space between the `Display legend` option toggle and the `Legend type` option in SV.

Before:
Screenshot 2020-09-22 at 14 58 46

After:
Screenshot 2020-09-22 at 14 58 57


packages/app/src/components/VisualizationOptions/Options/Legend.js
```diff
@@ -81,7 +81,12 @@ const Legend = ({
                     <div>
                         <FieldSet>
                             <UiCoreLegend>
-                                <span className={tabSectionTitle.className}>
+                                <span
+                                    className={tabSectionTitle.className}
+                                    style={{
+                                        marginTop: hideStyleOptions ? 8 : 0,
```
martinkrulltott, line 87: Use `spacers.dp8` instead of `8`?
Could all of this be put in a class that's used conditionally instead of using `style`?

edoardo: Refactored to use a class name with `cx` and a specific style for the additional margin.

martinkrulltott (approved):
Awesome 🚀

PR #1315, opened 2020-09-25
fix: assigned categories has an empty tooltip

martinkrulltott:
Issue: On hovering the chip for _Assigned categories_ it displays an empty tooltip. 

Cause: There is a check to only generate tooltip content for non-AC dimensions, however the `Tooltip` still renders even if the `content` is not provided.

image

Solution: Move the conditional rendering outside of the `Tooltip` component, to only render this component for non-AC dimensions. AC will instead get a "plain" chip without the `Tooltip` component. I added `renderChipContent` to decrease duplication and make sure the content itself is always the same.


edoardo (approved):
LGTM

PR #1334, opened 2020-10-07
refactor: rewrite visualization configs (TECH-433)

martinkrulltott:
Solves [TECH-433](https://jira.dhis2.org/browse/TECH-433)

Pure refactor change, no visual changes in the ui. Except:

1. `YoY Line` no longer have the _"No space between bars/column"_ option, since this was incorrect.

Previously, the concept of `sections` were used to create reusable parts to be included across different configs.
This has been extended with two new concepts, resulting in:

- `tabs`: a tab component, either self-contained (e.g. `series` and `legend`) or an empty shell (e.g. `data`) that requires the `content` param to be passed in.
- `templates`: an empty section component, that requires the `content` param to be passed in
- `sections`: a self-contained section component with default content. Doesn't allow `content` to be passed in (if customisation is required, `templates` should be used instead).

As per 089703942fb4729c75986014107bb6291c1f3fd9 most configs are replaced by a `defaultConfig` with optional params to conditionally toggle sections and content. Going back to this commit to see the old configs before they were deleted might make it easier to review this PR.

----------

**TODO:** 
- [x] Simplify configs and extract parts to be reused
- [ ] Add helper function for [hasCustomAxes logic](https://github.com/dhis2/data-visualizer-app/pull/1280/files#diff-3db980ee16ade99ecb3d3a7c6c9cbfdbR81)
- [x] Rename `hasCustomAxes` to something more generic (e.g. `hasDisabledSections`)
- [x] Reduce the number of configs by passing params to a default config instead
- [x] `YoY Line` shouldn't display _"No space between bars/column"_
- [x] Test and compare every tab for every vis type with master to verify this change has no unexpected visual impact.


martinkrulltott:
I'm skipping the _Add helper function for hasCustomAxes logic_ task as I can't find a clear pattern for how this should work in DV compared to Analytics. `isDualAxisType`, `hasCustomAxes` and `hasRelativeItems` are used together in DV but in Analytics they're also used individually, so the proposed helper function would only be used by DV in a single place, which I don't think requires a central helper function.

martinkrulltott:
I've gone through all tabs for all vis types and compared with master (debug/2.35.0) to make sure no unexpected visual changes were introduced. The above listed _YoY Line no longer have the "No space between bars/column" option_ was the only change I found.

jenniferarnesen:
Looks good. I briefly tried the options and things look the same.  A few observations:
- would it make sense to move `templates/` into `sections/` since templates are a custom section?
- I can't find anywhere in the code where `hasDisabledSections` is explained.  Maybe there should be some jsdoc for it.
- There are two functions named `getOptionsByType` exported from 2 different modules. Could this lead to confusion?

jenniferarnesen (approved):

PR #1355, opened 2020-10-19
feat: enable legend tab for column and bar (DHIS2-147)

martinkrulltott:
Implements [DHIS2-147](https://jira.dhis2.org/browse/DHIS2-147)

**Requires https://github.com/dhis2/analytics/pull/650 and https://github.com/dhis2/analytics/pull/656**

---

### Key features

1. Enable the `Legend` tab for `Column` and `Bar`
2. Refactor `config` to use props from Analytics instead of hardcoding them
3. Fetch `LegendSet` names from the API

---

### Description

Started out as a tiny change to enable the `Legend` tab for `Column` and `Bar`, the rest of the legend implementation can be found in the Analytics PR.

To prevent the `config` from mismatching with the Highcharts output, I refactored the `config` to use centralised props from Analytics instead of using hardcoded values. Meaning that the config will need minimal maintenance to work with future vis types.

`LegendSet` names are now fetched from the API, which are passed to Analytics to be used in the hover tooltip in Highcharts.

---

### Screenshots

_Legend tab enabled for Column_
image

_Legend tab enabled for Bar_
image


janhenrikoverland (approved):

packages/app/src/modules/options/config.js
```diff
@@ -1,18 +1,13 @@
...
-    VIS_TYPE_LINE,
-    VIS_TYPE_RADAR,
-    VIS_TYPE_AREA,
-    VIS_TYPE_STACKED_AREA,
     VIS_TYPE_PIE,
     VIS_TYPE_GAUGE,
-    VIS_TYPE_YEAR_OVER_YEAR_LINE,
-    VIS_TYPE_YEAR_OVER_YEAR_COLUMN,
     VIS_TYPE_SINGLE_VALUE,
+    isStacked as getIsStacked,
```
edoardo, line 6: Why the rename?
Perhaps it should be called `isStackedType` for consistency with the others?

edoardo (approved):

packages/app/src/modules/options/config.js
martinkrulltott: Because I wanted to pass `isStacked` to `defaultProps` without renaming the prop and obviously I can't have `const isStacked = isStacked(type)` 🙂  alternatively I could keep the original name of the Analytics prop and use a different name for my const, but I thought `isStacked: someOtherName` would be less clean the current way.

martinkrulltott: The exported prop from Analytics could be renamed on the Analytics side but that would cause a breaking change that I don't think is worth it.

edoardo: Yes, I thought about the breaking change.
Let's keep it as it is, or use `isStacked as isStackedType` 😆

PR #1360, opened 2020-10-20
refactor: get adapted layout from analytics

jenniferarnesen:
Related to [DHIS2-9599](https://jira.dhis2.org/browse/DHIS2-9599)

**Requires https://github.com/dhis2/analytics/pull/654**
**Note: build will fail until the change in analytics is released

There should be no functional or visual changes. The getLayout functionality is being moved to analytics to make it usable for other apps (like Dashboards, which needs it)

The implementation in analytics is modified to expose a single function, instead of separate functions for different vis types. The single function accepts the visualization type to provide the layout for, as well as the current layout


janhenrikoverland (approved):
LGTM

PR #1361, opened 2020-10-20
refactor: replace d2-ui-file-menu with analytics/FileMenu

edoardo:
Implements [DHIS2-9821](https://jira.dhis2.org/browse/DHIS2-9821)

**Requires https://github.com/dhis2/analytics/pull/655**
**Requires https://github.com/dhis2/analytics/pull/798**

---

### Key features

1. Replace `d2-ui-file-menu` with `FileMenu` component from `@dhis2/analytics`

---

### Description

There are no functional differences, only small style differences.
Refer to the analytics PR for more info and screenshot.

---

### Known issues

The use of Esc for closing the menu is not supported in `ui`.
Ideally we should keep the functionality, it would be helpful if it can be implemented in a centralized way, but given the open/close state is handled in the app via conditional rendering, I'm not sure how that can work.
See: https://github.com/dhis2/notes/discussions/215

**Solved** for now in the FileMenu itself.

### DIFF
--- a/cypress/elements/fileMenu/index.js
+++ b/cypress/elements/fileMenu/index.js
@@ -15,12 +15,13 @@ export const FILE_MENU_BUTTON_GETLINK = 'Get link…'
 export const FILE_MENU_BUTTON_DELETE = 'Delete'
 
 export const closeFileMenu = () =>
-    cy
+    cy.get('[data-test="file-menu-toggle-layer"]').click('topLeft')
+/*    cy
         .getBySel(fileMenuItemEl)
         .find('li')
         .contains(FILE_MENU_BUTTON_NEW)
         .type('{esc}', { force: true })
-
+*/
 export const clickFileMenuButton = buttonName =>
     cy.getBySel(fileMenuItemEl).find('li').contains(buttonName).click()
 
--- a/cypress/elements/fileMenu/index.js
+++ b/cypress/elements/fileMenu/index.js
@@ -1,7 +1,7 @@
 import { clickMenuBarFileButton } from '../menuBar'
 
-const deleteModalEl = '[data-test="file-menu-delete-modal"]'
-const fileMenuItemEl = '[data-test="file-menu-container"] li'
+const deleteModalEl = 'file-menu-delete-modal'
+const fileMenuItemEl = 'file-menu-container'
 
 export const FILE_MENU_BUTTON_NEW = 'New'
 export const FILE_MENU_BUTTON_OPEN = 'Open…'
@@ -16,22 +16,29 @@ export const FILE_MENU_BUTTON_DELETE = 'Delete'
 
 export const closeFileMenu = () =>
     cy
-        .get(fileMenuItemEl)
+        .getBySel(fileMenuItemEl)
+        .find('li')
         .contains(FILE_MENU_BUTTON_NEW)
         .type('{esc}', { force: true })
 
 export const clickFileMenuButton = buttonName =>
-    cy.get(fileMenuItemEl).contains(buttonName).click()
+    cy.getBySel(fileMenuItemEl).find('li').contains(buttonName).click()
 
 export const createNewAO = () => {
     clickMenuBarFileButton()
-    cy.get(fileMenuItemEl).contains(FILE_MENU_BUTTON_NEW).click()
+    cy.getBySel(fileMenuItemEl)
+        .find('li')
+        .contains(FILE_MENU_BUTTON_NEW)
+        .click()
 }
 
 export const deleteAO = () => {
     clickMenuBarFileButton()
-    cy.get(fileMenuItemEl).contains(FILE_MENU_BUTTON_DELETE).click()
-    cy.get(deleteModalEl)
+    cy.getBySel(fileMenuItemEl)
+        .find('li')
+        .contains(FILE_MENU_BUTTON_DELETE)
+        .click()
+    cy.getBySel(deleteModalEl)
         .find('button')
         .contains(FILE_MENU_BUTTON_DELETE)
         .click()
@@ -39,7 +46,8 @@ export const deleteAO = () => {
 
 export const expectFileMenuButtonToBeDisabled = (buttonName, inverse) =>
     cy
-        .get(fileMenuItemEl)
+        .getBySel(fileMenuItemEl)
+        .find('li')
         .contains(buttonName)
         .parents('li')
         .invoke('attr', 'class')
--- a/cypress/elements/fileMenu/index.js
+++ b/cypress/elements/fileMenu/index.js
@@ -7,25 +7,27 @@
 
 import { clickMenuBarFileButton } from '../menuBar'
 
-const deleteModalEl = '*[class^="MuiDialog-container"]'
-const fileMenuItemEl = '*[role="menuitem"]'
+const deleteModalEl = '[data-test="file-menu-delete-modal"]'
+const fileMenuItemEl = '[data-test="file-menu-container"] li'
 
 export const FILE_MENU_BUTTON_NEW = 'New'
-export const FILE_MENU_BUTTON_OPEN = 'Open'
-export const FILE_MENU_BUTTON_SAVE = 'Save'
-export const FILE_MENU_BUTTON_SAVEAS = 'Save as...'
-export const FILE_MENU_BUTTON_RENAME = 'Rename'
-export const FILE_MENU_BUTTON_TRANSLATE = 'Translate'
-export const FILE_MENU_BUTTON_SHARE = 'Share'
-export const FILE_MENU_BUTTON_GETLINK = 'Get link'
+export const FILE_MENU_BUTTON_OPEN = 'Open…'
+export const FILE_MENU_BUTTON_SAVE_EXISTING = 'Save'
+export const FILE_MENU_BUTTON_SAVE_NEW = 'Save…'
+export const FILE_MENU_BUTTON_SAVEAS = 'Save as…'
+export const FILE_MENU_BUTTON_RENAME = 'Rename…'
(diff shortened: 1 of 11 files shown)


cypress/elements/fileMenu/index.js
```diff
@@ -1,43 +1,31 @@
...
-    of the FileMenu being old and still uses MUI.
-    This can be refactored to use proper data-test selectors once the new
-    FileMenu component is in place.
-*/
-
 import { clickMenuBarFileButton } from '../menuBar'
 
-const deleteModalEl = '*[class^="MuiDialog-container"]'
-const fileMenuItemEl = '*[role="menuitem"]'
+const deleteModalEl = '[data-test="file-menu-delete-modal"]'
```
martinkrulltott, line 3: When you want to use a `data-test` prop, you just have to provide the name in the const and the use the `cy.getBySel` instead of `cy.get`.

cypress/elements/fileMenu/index.js
```diff
@@ -1,43 +1,31 @@
...
+const fileMenuItemEl = '[data-test="file-menu-container"] li'
```
martinkrulltott, line 4: The `li` part is not ideal and breaks the pattern that's used for all other elements. I suggest skipping the `li` here by setting `const fileMenuItemEl = 'file-menu-container` and using `cy.getBySel(fileMenuItemEl).find('li')`instead.

cypress/elements/fileMenu/index.js
martinkrulltott: The same goes for all other `'[data-test="XYZ"]'` below

cypress/elements/fileMenu/save.js
```diff
@@ -1,24 +1,25 @@
...
-const saveModalEl = '*[class^="MuiDialog-container"]' // TODO: Add data-test to save modal to target this better
-const saveModalSaveButtonEl = '[type="submit"]'
+const saveModalEl = '[data-test="file-menu-saveas-modal"]'
+const saveModalSaveButtonEl = '[data-test="file-menu-saveas-modal-save"]'
 
 export const saveNewAO = (name, description) => {
     clickMenuBarFileButton()
-    clickFileMenuButton(FILE_MENU_BUTTON_SAVE)
+    clickFileMenuButton(FILE_MENU_BUTTON_SAVE_NEW)
     cy.get(saveModalEl).find('input').clear().type(name)
```
martinkrulltott, line 15: Change to get the `file-menu-saveas-modal-name` instead.
If you don't want to use `getBySel(X)` directly only you can chain it like `cy.get(saveModalEl).findBySel(X)`.
Same goes for the textarea on the next line.

cypress/elements/fileMenu/save.js
```diff
@@ -28,7 +29,7 @@ export const saveAOAs = (name, description) => {
         cy.get(saveModalEl).find('input').clear().type(name)
```
martinkrulltott, line 29: Same as the comment above

packages/app/src/components/MenuBar/MenuBar.js
```diff
@@ -32,7 +32,8 @@ const onNew = () => {
     }
 }
 const getOnRename = props => details => props.onRenameVisualization(details)
-const getOnSave = props => details => props.onSaveVisualization(details, false)
+const getOnSave = props => (details = {}) =>
+    props.onSaveVisualization(details, false)
```
martinkrulltott, line 36: Why not leave the the default prop handling to `onSaveVisualization` instead of using a `details = {}` here?

cypress/elements/fileMenu/index.js
```diff
@@ -1,57 +1,54 @@
...
-    This can be refactored to use proper data-test selectors once the new
-    FileMenu component is in place.
-*/
-
 import { clickMenuBarFileButton } from '../menuBar'
 
-const deleteModalEl = '*[class^="MuiDialog-container"]'
-const fileMenuItemEl = '*[role="menuitem"]'
+const deleteModalEl = 'file-menu-delete-modal'
+const fileMenuItemEl = 'file-menu-container'
```
martinkrulltott, line 4: If the content is `file-menu-container`, then this should probably be renamed to `fileMenuContainerEl`

cypress/elements/fileMenu/index.js
```diff
@@ -1,57 +1,54 @@
...
+export const FILE_MENU_BUTTON_RENAME = 'Rename…'
+export const FILE_MENU_BUTTON_TRANSLATE = 'Translate…'
+export const FILE_MENU_BUTTON_SHARE = 'Share…'
+export const FILE_MENU_BUTTON_GETLINK = 'Get link…'
 export const FILE_MENU_BUTTON_DELETE = 'Delete'
 
 export const closeFileMenu = () =>
-    cy
-        .get(fileMenuItemEl)
+    cy.get('[data-test="file-menu-toggle-layer"]').click('topLeft')
```
martinkrulltott, line 18: Why didn't this get a `const` at the top of the file and why not use `getBySel`?

cypress/elements/fileMenu/index.js
```diff
@@ -1,57 +1,54 @@
...
-        .get(fileMenuItemEl) // TODO: Change once new FileMenu is in place
-        .contains(buttonName)
-        .click()
+    cy.getBySel(fileMenuItemEl).find('li').contains(buttonName).click()
 
 export const createNewAO = () => {
     clickMenuBarFileButton()
-    cy.get(fileMenuItemEl) // TODO: Change once new FileMenu is in place
+    cy.getBySel(fileMenuItemEl)
+        .find('li')
```
martinkrulltott, line 31: It looks to me like `cy.getBySel(fileMenuItemEl).contains(...)` would be enough, using `find('li')` seems redundant. But maybe I'm missing something?

cypress/elements/fileMenu/open.js
```diff
@@ -5,7 +5,7 @@ import { clickMenuBarFileButton } from '../menuBar'
 const openModalEl = '*[class^="MuiDialogContent"]' // TODO: Add data-test to open modal to target this better
 const openModalFooterEl = '*[class^="MuiTableFooter"]'
 const openModalToolbarEl = '*[class^="MuiToolbar-root"]'
-const createdByOthersEl = '[data-value="byothers"]'
+const createdByOthersEl = 'byothers'
```
martinkrulltott, line 8: This won't work. `getBySel` only works with `data-test` props. So ideally this element should be given a `data-test` prop instead. But we're still using the old favourites dialog right, so if it's very tedious to add maybe leave it as it was before?

martinkrulltott: Also, if everything still passes even though this test isn't working, we should add a check to make sure that the `searchAOByName` actually returned the right thing...

martinkrulltott: Or did you actually add `data-test="byothers"` on the element? If so the prop should probably be renamed to better follow our naming convention.

martinkrulltott (commented):
_See individual comments_

cypress/elements/fileMenu/index.js
edoardo: Indeed it seems enough, although now is using `fileMenuContainerEl` renamed selector.

cypress/elements/fileMenu/open.js
edoardo: Reverted back. `data-value` is used in the component itself, so changing it will break the component.
The favourite dialog is going to be rewritten eventually, so I will leave it for now.

cypress/elements/fileMenu/index.js
edoardo: Fixed.

cypress/elements/fileMenu/index.js
edoardo: Done.

packages/app/src/components/MenuBar/MenuBar.js
edoardo: Sorted.

cypress/elements/fileMenu/index.js
edoardo: Done.

cypress/elements/fileMenu/index.js
edoardo: Removed the find for `li` as discussed in the other comment.

edoardo: Done.

cypress/elements/fileMenu/save.js
edoardo: Done.

cypress/elements/fileMenu/save.js
edoardo: Done.

cypress/integration/start.cy.js
```diff
@@ -78,12 +79,12 @@ describe('viewing the start screen', () => {
...
             FILE_MENU_BUTTON_NEW,
             FILE_MENU_BUTTON_OPEN,
-            FILE_MENU_BUTTON_SAVE,
+            FILE_MENU_BUTTON_SAVE_NEW,
         ]
         enabledButtons.forEach(button =>
             expectFileMenuButtonToBeEnabled(button)
         )
-        closeFileMenu()
+        closeFileMenuWithEsc()
```
martinkrulltott, line 87: No biggie, but I would consider wrapping this in it's own `it` (example below) or renaming the test to `primary File menu buttons are enabled and close with esc key` to make it obvious for us in the future to figure out why this particular case failed if something was wrong with the esc key press in particular.

> it('File menu is closed with escape', () => {
             closeFileMenuWithEsc()
       })

Same goes for `closeFileMenuWithClick` below.

martinkrulltott (approved):

cypress/integration/start.cy.js
edoardo: Done

PR #1369, opened 2020-10-21
refactor: convert d2 to app runtime for api CRUD operations

edoardo:
**Requires https://github.com/dhis2/analytics/pull/690**

---

### Key features

1. Replace `d2` with `app-runtime` for api CRUD operations

---

### Description

This is required for leveraging data caching and other upcoming features in `app-runtime`.
`d2` is still in use in various part of DV.
In some places the conversion is relatively easy, in some others not and we have to figure out how to tackle those (ie. use data store,`d2.analytics` ...).

---

### TODO

-   ~[ ] figure out what to do with user data store~ (**will be addressed in a separate PR**)

---


packages/app/src/D2Shim.js
```diff
@@ -6,7 +6,6 @@ import { useState } from 'react'
 import { init as d2Init, config as d2Config, getUserSettings } from 'd2'
```
jenniferarnesen, line 6: Not sure, but this could be the right time to remove this and use the `app-runtime-adapter-d2`? There are some things in this file that are not in that lib though: `history`, and `extractUserSettings` so those would have to be done elsewhere

packages/plugin/src/VisualizationPlugin.js
```diff
@@ -65,23 +62,31 @@ export const VisualizationPlugin = ({
...
 
+    useEffect(() => {
+        const doFetchOuLevels = async () => {
+            const ouLevels = await apiFetchOrganisationUnitLevels(engine)
+
+            setOuLevels(ouLevels)
+        }
+
+        doFetchOuLevels()
+    }, [engine])
```
jenniferarnesen, line 88: Does the engine change? Does it have to be in the array, or does it work as expected with an empty array? (Not sure if this would be preferred over the current, actually)

packages/app/src/api/settings.js
```diff
@@ -1,22 +1,20 @@
...
+        key: SYSTEM_SETTINGS,
+    },
 }
 
-export const apiFetchUserSettings = () => {
-    // "key" params not supported in the backend yet
-    // update: now it is, should be used here
-    const endPoint = '/userSettings'
+export const apiFetchSystemSettings = async dataEngine => {
+    const systemSettingsData = await dataEngine.query(
```
jenniferarnesen, line 12: Very minor - should we be consistent in calling it "engine" or "dataEngine"? (see api/mostViewedVisualizations)

jenniferarnesen (approved):

jenniferarnesen:
Approved - make sure checks are passing before merge.

packages/plugin/src/VisualizationPlugin.js
edoardo: It shouldn't change, but having it avoids the linting errors.

packages/app/src/api/settings.js
edoardo: Yes, i thought `dataEngine` is more descriptive, as we might have other engines in the future (for example we have the pivot table engine).
I will rename the other one.

packages/app/src/D2Shim.js
edoardo: I'm going to look into it.

amcgee: Those should be able to be done in the callbacks in the lib, I believe

edoardo: I started removing this shim in favour of the one from the adapter.
I'll push the changes soon.

edoardo: Will add these changes in a separate PR.
This one is growing too big.

martinkrulltott (approved):

PR #1381, opened 2020-10-23
fix: prevent options from crashing by filtering out empty objects (DHIS2-147)

martinkrulltott:
Fixes [DHIS2-147](https://jira.dhis2.org/browse/DHIS2-147)

---

### Description

The Options modal is crashing for all visTypes that don't support legends because `supportsLegends ? getLegendTab({ hideStyleOptions: true }) : {}` ([link](https://github.com/dhis2/data-visualizer-app/blob/8feb92f1f8f7724e0905f35719af2b1490f1ed1a/packages/app/src/modules/options/defaultConfig.js#L28)) leaves an empty object behind that is then used in the options modal generation (which results in `{}.content.map(...)` which crashes.
To prevent this, I changed the condition to return `false` and then filter the resulting array to get rid of the `false`. The end result is an array that only has proper objects.


edoardo (approved):
👍 
I'm just not sure about the comment, it doesn't really help *if* you spent the time to understand what is actually happening with the filter.

PR #1412, opened 2020-10-30
test: Improve Cypress tests

martinkrulltott:
**Requires https://github.com/dhis2/analytics/pull/679**

---

### Key features

1. Cypress testing
2. CI integration for automatic test runs on [dashboard.cypress.io](https://dashboard.cypress.io/projects/sojh88/)

---

### Goals

1. Support static baseline / "happy path" testing for each feature in isolation (**this PR**)
2. Support dynamic combination testing, pulling in a new feature into a pre-made combination (where the new feature makes sense)
3. Support random smart monkey testing of combinations to see if anything can cause the app to crash or throw errors to the console
4. Use snapshot testing for 1. (and some of 2.) to verify the visual output of the Highcharts lib

---

### Out of scope

1. Testing individual output values (i.e. the result from the backend)
2. Testing the output of individual features in monkey tests (e.g. adding a trendline to Gauge counts as a pass even though it's not visible, as long as the app doesn't crash / throw errors).

---

### Description

Improve Cypress testing in DV. Base line tests for the following parts / components:

- View the start page
- Create a new visualization
- Save
- Load / open
- Interacting with the dimensions panel
- Interacting with the layout section
- Using the "confirm leave" modal
- Options (WIP)

---

### TODO

-   [ ] Add condition to only run `onConfigGenerated` in dev mode, not in production


cypress.env.json
```diff
@@ -0,0 +1,6 @@
+{
```
jenniferarnesen, line 1: We don't typically commit developer env files, right? I would instead add instructions in the readme for the needed env setup, possibly including a `.example` config file.  Also, `dhis2_datatest_prefix` is not in use.

jenniferarnesen:
Could you update the README so that developers can easily configure and run the tests locally?

cypress/assets/unfetch.umd.js
```diff
@@ -0,0 +1,83 @@
+/**
```
jenniferarnesen, line 1: Is this file in use? I think it is only used if the tests are running against fixtures. If not in use, then remove it (you can always re-add it later from one of the other repos. Or more likely, it will probably be part of cli-utils-cypress in the future)

cypress/fixtures/network/requests.json
```diff
@@ -0,0 +1,42 @@
+{
```
jenniferarnesen, line 1: If this file is not in use, then remove.

cypress/support/index.js
```diff
@@ -0,0 +1,17 @@
+import '@dhis2/cli-utils-cypress/support'
+import './commands'
+
+import { loginAndPersistSession } from './server'
+
+//enableNetworkShim()
```
jenniferarnesen, line 6: Remove all commented-out code in the file

cypress/support/server/Networkshim.js
```diff
@@ -0,0 +1,151 @@
+import { getFileName } from './utils.js'
```
jenniferarnesen, line 1: Remove all code related to the network shim, since it isn't in use and may never be.

package.json
```diff
@@ -25,7 +25,13 @@
         "test": "yarn test-plugin && yarn test-app",
         "coverage": "yarn test-plugin --coverage && yarn test-app --coverage",
         "validate-commit": "d2-style js check --staged && d2-style text check --staged",
-        "validate-push": "yarn test"
+        "validate-push": "yarn test",
+        "cypress:start": "BROWSER=none yarn start",
+        "cy:open-live": "cypress_dhis2_api_stub_mode=DISABLED d2-utils-cypress open --appStart 'yarn cypress:start'",
```
jenniferarnesen, line 30: The test set up doesn't look at the `cypress_dhis2_api_stub_mode`, right? Then these can be removed from the scripts. Also, `cy:capture` can be removed completely.

packages/app/src/reducers/config.js
```diff
@@ -0,0 +1,15 @@
...
+    switch (action.type) {
+        case SET_CONFIG: {
+            return action.value
+        }
+        default:
+            return state
+    }
+}
+
+export const sGetVisConfig = state => state.visConfig
```
jenniferarnesen, line 15: This selector doesn't seem to be used anywhere.

cypress/assets/unfetch.umd.js
martinkrulltott: It's used in `server/visitWithUnfetchOverwriteFn`, which in turn overwrites the `visit` command. I don't know enough about fixtures to make this call. It's not using fixtures now, but will it ever?

cypress/support/server/Networkshim.js
martinkrulltott: does that mean removing the whole `server/enableNetworkShim.js` file as well then?

packages/app/src/reducers/config.js
martinkrulltott: True, since cypress reads the store through `getReduxState` rather than selectors

martinkrulltott: Removed

package.json
martinkrulltott: Removed and renamed the remaining two scripts to just `cy:open` and `cy:run`

cypress/support/index.js
martinkrulltott: Removed

cypress/fixtures/network/requests.json
martinkrulltott: Removed

cypress.env.json
martinkrulltott: Removed and added to list in `gitignore`

cypress/integration/save.cy.js
```diff
@@ -0,0 +1,188 @@
...
+const TEST_VIS_NAME = `TEST ${new Date().toLocaleString()}`
+const TEST_VIS_NAME_UPDATED = `${TEST_VIS_NAME} - updated`
+const TEST_VIS_DESCRIPTION = 'Generated by Cypress'
+const TEST_VIS_TYPE = getRandomVisType()
+const TEST_VIS_TYPE_NAME = visTypeDisplayNames[TEST_VIS_TYPE]
+
+// TODO: Add test to check that the description is saved and shown in the interpretations panel
+
+describe('saving an AO', () => {
+    describe('"save" and "save as" for a new AO', () => {
```
jenniferarnesen, line 60: These tests seem to break some Cypress best practices rules:
* https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests
* https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests

I tend to agree with the suggestions in these links. I think this would be better as a single test. As it stands now, they cannot be run in isolation.

jenniferarnesen:
The file names in `/elements` are inconsistent wrt capitalization. Since these files don't export classes, then I would rather go for starting with lowercase letter.

jenniferarnesen:
Suggest removing commented-out code in all the files.

README.md
```diff
@@ -30,35 +30,29 @@ Run unit tests with coverage report
...
-```
-
-There are plans to make this configurable. In order to run the tests, you need to define three environment variables:
-
-```
-CYPRESS_LOGIN_URL=http://localhost:8080 # This must match DHIS2_BASE_URL
-CYPRESS_DHIS2_USERNAME=myusername
-CYPRESS_DHIS2_PASSWORD=mypassword
+{
+    "dhis2_base_url": "https://debug.dhis2.org/dev",
```
jenniferarnesen, line 39: Sounds like Austin wanted to keep this server private-ish. Suggest changing this to localhost:8080

README.md
```diff
@@ -30,35 +30,29 @@ Run unit tests with coverage report
...
+    "dhis2_username": "admin",
+    "dhis2_password": "district",
+    "dhis2_datatest_prefix": "dhis2-datavisualizer"
```
jenniferarnesen, line 42: Remove this line

README.md
martinkrulltott: Fair enough, keep in mind that the debug url is still listed further down on the readme though: https://github.com/dhis2/data-visualizer-app/commit/7c3dd5aa75248a58c5a974e932db57218a28a707#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L92, thoughts @amcgee ?

cypress/integration/save.cy.js
```diff
@@ -0,0 +1,188 @@
...
+            expectAOTitleToBeUnsaved()
+            expectRouteToBeEmpty()
+        })
+        it('checks that Save as is disabled', () => {
+            clickMenuBarFileButton()
+            expectFileMenuButtonToBeDisabled(FILE_MENU_BUTTON_SAVEAS)
+            closeFileMenu()
+        })
+        it('saves new AO using "Save"', () => {
+            saveNewAO(TEST_VIS_NAME, TEST_VIS_DESCRIPTION)
```
jenniferarnesen, line 85: Is this AO ever deleted?

cypress/integration/layout.cy.js
```diff
@@ -0,0 +1,103 @@
...
+            describe(`${dim.name}`, () => {
+                it('clicks the dimension', () => {
+                    openDimension(dim.id)
+                })
+                it('shows the dimension modal', () => {
+                    expectDimensionModalToBeVisible(dim.id)
+                })
+                it('clicks the modal hide button', () => {
+                    clickDimensionModalHideButton()
+                })
```
jenniferarnesen, line 48: To me this is a good candidate for consolidating into a single test "Data dimension modal opens and closes".

Also, these tests have a duplicated assertion `expectDimensionModalToBeVisible` (check `openDimension` function)

cypress/integration/layout.cy.js
```diff
@@ -0,0 +1,103 @@
...
+        )
+        describe(`${TEST_DIM.name}`, () => {
+            it('opens the context menu', () => {
+                openContextMenu(TEST_DIM.id)
+            })
+            it(`moves to ${TEST_AXIS} axis`, () => {
+                clickContextMenuMove(TEST_DIM.id, TEST_AXIS)
+            })
+            it(`${TEST_AXIS} axis has dimension`, () => {
+                expectAxisToHaveDimension(TEST_AXIS, TEST_DIM.id)
```
jenniferarnesen, line 65: Consider consolidating here too

cypress/integration/layout.cy.js
```diff
@@ -0,0 +1,103 @@
...
+        describe(`${TEST_DIM.name}`, () => {
+            it('opens the context menu', () => {
+                openContextMenu(TEST_DIM.id)
+            })
+            it(`removes ${TEST_DIM.name}`, () => {
+                clickContextMenuRemove(TEST_DIM.id)
+            })
+            it(`${TEST_DIM.name} is removed`, () => {
+                expectLayoutToNotHaveDimension(TEST_DIM.id)
+            })
```
jenniferarnesen, line 82: And here...

cypress/integration/layout.cy.js
```diff
@@ -0,0 +1,103 @@
...
+            clickContextMenuDimSubMenu(TEST_DIM.id)
+            clickContextMenuAdd(TEST_DIM.id, TEST_AXIS)
+            expectDimensionToHaveSelectedStyle(TEST_DIM.id)
+            expectAxisToHaveDimension(TEST_AXIS, TEST_DIM.id)
+        })
+        it(`removes ${TEST_DIM.name}`, () => {
+            openContextMenu(DIMENSION_ID_DATA)
+            clickContextMenuRemove(TEST_DIM.id)
+            expectAxisToNotHaveDimension(TEST_AXIS, TEST_DIM.id)
+        })
```
jenniferarnesen, line 101: Consider here as well: "Add and removes Assigned Categories to the layout"

cypress/elements/DimensionsPanel.js
```diff
@@ -0,0 +1,92 @@
...
+
+const getDimensionButtonById = dimensionId => `${dimButtonEl}-${dimensionId}`
+
+export const openContextMenu = dimensionId =>
+    cy.getBySel(`${dimContextMenuButtonEl}-${dimensionId}`).click()
+
+export const openDimension = dimensionId => {
+    cy.getBySel(getDimensionButtonById(dimensionId)).click()
+    expectDimensionModalToBeVisible(dimensionId)
+}
```
jenniferarnesen, line 25: These functions have similar names, but one does an assertion and the other doesn't. Maybe this should be consistent?

cypress/integration/save.cy.js
martinkrulltott: Yeah its operating on the same AO the whole time. First it does a "Save" of the unsaved state, then a regular "Save" of the updates, then a "Save as" which renames it, then a "Delete". This breaks the test interdependency best practice but in turn also makes sure that tests are self-cleaning.

cypress/integration/layout.cy.js
martinkrulltott: Yes I agree. Initially I thought it would be good if each action could self-evaluate (e.g. `openDimension` checking if dimension is open), but I realised it led to a lot of duplication. I'll go through the earlier tests and refactor this.

The line 39 `describe` acts as a wrapper to get a list of dimension names in the test tree structure, but I guess suffixing the description with the dimension name, like `clicking layout items - Data` achieves the same thing with less code.

cypress/elements/DimensionsPanel.js
martinkrulltott: Yeah, I agree, too much duplication, as mentioned here as well https://github.com/dhis2/data-visualizer-app/pull/1412/#discussion_r524959611. Will refactor this now.

cypress/integration/options/verticalAxis.cy.js
```diff
@@ -0,0 +1,105 @@
...
+    setVerticalAxisRangeMinValue,
+    setVerticalAxisTitle,
+} from '../../elements/OptionsModal'
+import {
+    expectStoreConfigYAxisToHaveRangeMinValue,
+    expectStoreConfigYAxisToHaveRangeMaxValue,
+    expectStoreConfigYAxisToHaveTitleText,
+} from '../../utils/store'
+
+const dimensionId = DIMENSION_ID_DATA
```
jenniferarnesen, line 30: Are you creating a new constant because in the future this test will loop through other dimensions? (Same question with the `optionsTab` const on line 32.

cypress/elements/Chart.js
```diff
@@ -0,0 +1,113 @@
...
+
+const visualizationContainerEl = 'visualization-container'
+const visualizationTitleEl = 'visualization-title'
+const visualizationSubtitleEl = 'visualization-subtitle'
+const chartContainerEl = '.highcharts-container'
+const highchartsLegendEl = '.highcharts-legend'
+const highchartsTitleEl = '.highcharts-title'
+//const higchartsLegendItemEl = '.highcharts-legend-item'
+const unsavedVisualizationTitleText = 'Unsaved visualization'
+const AOTitleEl = 'AO-title'
```
jenniferarnesen, line 19: Sometimes constants are AO_TITLE_EL, other times AOTitleEl. Maybe aim for consitency (I'm guilty of this too).

cypress/elements/DimensionModal/dataDimension.js
```diff
@@ -0,0 +1,81 @@
+import { DIMENSION_ID_DATA } from '@dhis2/analytics'
+import { expectDimensionModalToBeVisible } from '.'
+
+const unselectedListEl = 'data-dimension-item-selector-unselected-items-list'
+//const unselectedItemEl = 'data-dimension-item-selector-unselected-items-list-item'
+const dataTypesSelectButtonEl = 'data-dimension-data-types-select-field-content'
+//const selctedListEl = 'data-dimension-item-selector-selected-items-list'
```
jenniferarnesen, line 7: Suggest deleting unused code. Maybe you can store it in a separate branch for use later?

cypress/integration/options/verticalAxis.cy.js
martinkrulltott: Yeah,  I've tried to keep things as dynamic and flexible as possible. You'll find similar things in a lot of places where I've extracted parts that could be changed or looped over in the future and sometimes I also loop through arrays with only a single item.

cypress/integration/layout.cy.js
martinkrulltott: Fixed 👍

cypress/elements/Chart.js
martinkrulltott: I'm yet to work out a good rule for this pattern, but my thought was that constants that are just picked arbitrarily to run the test (e.g. a data element, a vis type or a name) use SCREAMING_SNAKE (prefixed with TEST_ to make it even more obvious), but anything that's "a regular" const (a return value, expected output, the current item in a for loop etc) use our regular camelCase. 
I find that it's important to separate arbitrary consts from the rest, since the fact that they were indeed just picked arbitrarily makes them available to be changed, made in to an array and be looped through etc.
However I've been quite inconsistent with this convention so Im updating it now.

martinkrulltott: the above applies to `/integration` files. in `/elements` our regular pattern of STRING_CONSTS and regularConsts should be used. However I camelCase and suffix all element names with `El`, not sure this is the best way though, open for suggestions 🙂

cypress/elements/DimensionModal/dataDimension.js
martinkrulltott: Yeah, good point. I've removed most of the commented code and only kept some that are WIP, but I added a description to those for future reference.

cypress/integration/save.cy.js
martinkrulltott: Indeed, very valid points. As mentioned in PMs I've now refactored most of these to better follow the best practice rules, however in some cases deviate to instead follow a `describe - describe - it` pattern (where the second `describe` acts like an `it` and the `it` is a subpart of a "test") to get a compromise between following the rules and having good usability / readability / ease of debugging for devs.

.prettierignore
```diff
@@ -3,4 +3,5 @@
 /packages/plugin/node_modules/*
 /packages/app/build/*
 /packages/plugin/build/*
-/coverage/*
\ No newline at end of file
+/coverage/*
+/cypress/fixtures
```
jenniferarnesen, line 7: newline

cypress/elements/layout.js
```diff
@@ -0,0 +1,119 @@
...
+export const openDimension = dimensionId =>
+    cy.getBySel(getDimensionChipEl(dimensionId)).click()
+
+export const openContextMenu = dimensionId =>
+    cy.getBySel(`${chipMenuButtonEl}-${dimensionId}`).click()
+
+export const clickContextMenuAdd = (dimensionId, axisId) =>
+    cy
+        .getBySel(`${chipMenuActionOptionEl}-${dimensionId}-to-${axisId}`)
+        .should('contain', 'Add to')
```
jenniferarnesen, line 87: Could this be `.contains('Add to').click()` instead? Since this is an action function, not an expectation function.  (same comment for other places in this file)

cypress/elements/optionsModal/verticalAxis.js
```diff
@@ -0,0 +1,50 @@
...
+const titleInputEl = 'option-vertical-axis-title-input'
+const rangeMinInputEl = 'option-vertical-axis-range-min-input'
+const rangeMaxInputEl = 'option-vertical-axis-range-max-input'
+
+export const enableVerticalAxisTitle = () =>
+    cy
+        .getBySel(titleCheckboxEl)
+        .find('[type="checkbox"]')
+        .check({ force: true })
+        .should('be.checked')
```
jenniferarnesen, line 11: Is this function actually an expectation? Then it should be renamed as such. But it seems like an action function, in which case I would argue that the `should` should be dropped.

cypress/integration/options/fontStyles.cy.js
```diff
@@ -0,0 +1,184 @@
...
+            expectStoreConfigTitleToBeValue(CONFIG_DEFAULT_TITLE)
+        })
+        it('opens Options -> Style', () => {
+            clickMenuBarOptionsButton()
+            clickOptionsTab(OPTIONS_TAB_STYLE)
+        })
+        it('changes the font size', () => {
+            changeFontSizeOption(type, TEST_FONT_SIZE_OPTION.label)
+        })
+        it('changes the font size', () => {
```
jenniferarnesen, line 102: Wrong test name?

cypress/integration/options/fontStyles.cy.js
```diff
@@ -0,0 +1,184 @@
...
+            clickMenuBarOptionsButton()
+            clickOptionsTab(OPTIONS_TAB_STYLE)
+        })
+        it('sets a custom subtitle', () => {
+            setCustomSubtitle(TEST_SUBTITLE_TEXT)
+        })
+        it('changes the font size', () => {
+            changeFontSizeOption(type, TEST_FONT_SIZE_OPTION.label)
+        })
+        it('changes the font size', () => {
```
jenniferarnesen, line 151: Wrong test name

cypress/integration/options/legend.cy.js
```diff
@@ -0,0 +1,12 @@
+/*  TODO:
```
jenniferarnesen, line 1: Instead of having test files that only have TODOs and no code, maybe there should be a jira epic with issues for each of these TODOS and FIXMEs?

cypress/integration/save.cy.js
```diff
@@ -0,0 +1,182 @@
...
+            expectAOTitleToBeValue(TEST_VIS_NAME)
+            expectVisualizationToBeVisible(TEST_VIS_TYPE)
+        })
+        it('checks that the url was changed', () => {
+            expectRouteToBeAOId()
+        })
+        it('all File menu buttons are enabled', () => {
```
jenniferarnesen, line 93: This doesn't actually check all the File menu buttons? (NEW, SAVE, OPEN missing)

cypress/integration/start.cy.js
```diff
@@ -0,0 +1,87 @@
...
+    it('period dimension has 1 item', () => {
+        expectDimensionToHaveItemAmount(DIMENSION_ID_PERIOD, 1)
+    })
+    it('axis filter has orgunit dimension', () => {
+        expectAxisToHaveDimension(AXIS_ID_FILTERS, DIMENSION_ID_ORGUNIT)
+    })
+    it('orgunit dimension has 1 item', () => {
+        expectDimensionToHaveItemAmount(DIMENSION_ID_ORGUNIT, 1)
+    })
+    it('File menu buttons are disabled', () => {
```
jenniferarnesen, line 72: Should it check that OPEN, NEW, SAVE are enabled?

cypress/support/server/index.js
```diff
@@ -0,0 +1 @@
+export { default as loginAndPersistSession } from './loginAndPersistSession.js'
```
jenniferarnesen, line 1: Seems unnecessary to have a /server directory now.

cypress/utils/data.js
```diff
@@ -0,0 +1,90 @@
...
+    VIS_TYPE_COLUMN,
+    VIS_TYPE_GAUGE,
+    VIS_TYPE_LINE,
+    VIS_TYPE_PIE,
+    VIS_TYPE_PIVOT_TABLE,
+    VIS_TYPE_STACKED_COLUMN,
+    VIS_TYPE_YEAR_OVER_YEAR_LINE,
+} from '@dhis2/analytics'
+
+//const getTestAOsFromEndpoint = () => {
```
jenniferarnesen, line 12: Add "TODO" ?

jenniferarnesen (approved):

cypress/elements/layout.js
martinkrulltott: This is actually an assertion to check that the option says `Add to` and not `Move to`, because the option label switches between "add" and "move" if the dimension is in the layout already or not. `contains` would also work and since its an action it might be better to not use an assertion here.

cypress/elements/optionsModal/verticalAxis.js
martinkrulltott: This function is a workaround to check the `dhis2/ui` `checkbox` components, because they're custom checkboxes which Cypress can't "see", which is why the `force: true` is used. However, to verify that the checkbox was actually checked (since it could be hidden for another reason) I added the should to be sure. I'll add a `FIXME` comment to find another way to check the checkbox though.

cypress/integration/options/fontStyles.cy.js
martinkrulltott: Yepp, classic copy/paste mistake

cypress/integration/options/legend.cy.js
martinkrulltott: Yeah would be good to plan what needs to be tested and refactored in a central place. @janhenrikoverland ?

martinkrulltott: We've discussed this and leaving the comments as is for now makes the most sense. We can start a proper JIRA epic once we pick up the holistic regression test writing again in the future. For now we'll only add continuous feature tests as we go.

PR #1477, opened 2020-11-24
test: replace period instead of data in confirmLeave to avoid flakiness

martinkrulltott:
### Key features

1. fix flaky test
2. add a common way to change an AO

---

### Description

`confirmLeave` is flaky, I think this is caused by the way MUI handles updates in the data dimension selector (soon to be replaced by non-MUI). The test replaces data items which sometimes isn't performed in order, even though a `then` is used. This PR changes the test to replace period items instead, since the new period selector (non-MUI) seems to be stable. 

`replacePeriodItems` was added to create a common way to change an AO that's easy to use. Period, Data and Org units are the only dimensions that can be assumed are in all AOs, so when loading a "random" AO that should be changed, it must change one of these three dimensions. However Data is flaky (issue above) and the org unit tree is quite complex (plus it might suffer from the same flakiness as Data), which is why I chose to change period instead. I've picked a period which I think is fairly uncommon (`Last 2 six-months`). To be able to alternate between this and an another value (`Quarters per year`) I've added a `useAltData` prop to switch between the two periods.

### DIFF
--- a/cypress/elements/common.js
+++ b/cypress/elements/common.js
@@ -1,0 +1,26 @@
+import { isYearOverYear, DIMENSION_ID_PERIOD } from '@dhis2/analytics'
+
+import { clickMenuBarUpdateButton } from './menuBar'
+import { openDimension, selectYoyCategoryOption } from './layout'
+import {
+    removeAllPeriodItems,
+    selectRelativePeriods,
+} from './dimensionModal/periodDimension'
+import { clickDimensionModalUpdateButton } from './dimensionModal'
+
+export const replacePeriodItems = (visType, useAltData) => {
+    if (isYearOverYear(visType)) {
+        const TEST_PERIOD = !useAltData
+            ? 'Last 2 six-months'
+            : 'Quarters per year'
+        selectYoyCategoryOption(TEST_PERIOD)
+        clickMenuBarUpdateButton()
+    } else {
+        const TEST_PERIOD_TYPE = !useAltData ? 'Six-months' : 'Quarters'
+        const TEST_PERIOD = !useAltData ? 'Last six-month' : 'Last quarter'
+        openDimension(DIMENSION_ID_PERIOD)
+        removeAllPeriodItems()
+        selectRelativePeriods([TEST_PERIOD], TEST_PERIOD_TYPE)
+        clickDimensionModalUpdateButton()
+    }
+}
--- a/cypress/elements/common.js
+++ b/cypress/elements/common.js
@@ -8,7 +8,11 @@ import {
 } from './dimensionModal/periodDimension'
 import { clickDimensionModalUpdateButton } from './dimensionModal'
 
-export const replacePeriodItems = (visType, useAltData) => {
+export const replacePeriodItems = (
+    visType,
+    options = { useAltData: false }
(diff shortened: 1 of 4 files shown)


cypress/integration/confirmLeave.cy.js
```diff
@@ -11,30 +12,42 @@ import {
...
-        removeAllDataItems().then(() => expectNoDataItemsToBeSelected())
-        selectDataElements([getRandomArrayItem(TEST_DATA_ELEMENTS).name])
-        clickDimensionModalUpdateButton()
+    it(`replaces the selected period`, () => {
+        if (isYearOverYear(TEST_AO.type)) {
+            const TEST_PERIOD = 'Last 2 six-months'
+            selectYoyCategoryOption(TEST_PERIOD)
+            clickMenuBarUpdateButton()
+        } else {
+            const TEST_PERIOD_TYPE = 'Six-months'
```
jenniferarnesen, line 43: Doesn't seem necessary to create consts for values that are only used once? (same with line 39)

jenniferarnesen (approved):

cypress/integration/confirmLeave.cy.js
martinkrulltott: Yeah good point, it was a copy/paste from a WIP section that I've now moved to a common file and finished 👍

cypress/elements/common.js
```diff
@@ -0,0 +1,26 @@
...
+
+import { clickMenuBarUpdateButton } from './menuBar'
+import { openDimension, selectYoyCategoryOption } from './layout'
+import {
+    removeAllPeriodItems,
+    selectRelativePeriods,
+} from './dimensionModal/periodDimension'
+import { clickDimensionModalUpdateButton } from './dimensionModal'
+
+export const replacePeriodItems = (visType, useAltData) => {
```
jenniferarnesen, line 11: `useAltData` is kind of an unclear arg to the function. What does it mean... Instead, I would go with defaults:

const replacePeriodItems = (visType, period)

//for non-yoy
const testPeriodType = period.type ? period.type : 'Six-months'
const testPeriod = period.name ? period.name : 'Last six-month'

//for yoy
const testPeriod = period.name ? period.name : 'Last 2 six-months'

This would also make the test more readable (`replacePeriodItems(visType, {name: 'Quarters per year'})` instead of `replacePeriodItems(visType, true))`

Actually, I might even argue that the period info should always be passed to the function instead of having defaults.

martinkrulltott: Yeah I would like to pass it too, but I have to pass 3 props then, as you need `nonYoyPeriodName`, `nonYoyPeriodType` and `yoyPeriodName`, since yoy periods aren't called the same as regular periods. So the keep it simple I thought I'd just provide two sets of the 3 props and just alternate between them. ATM I can't think of a use-case that would need more than two sets anyway (since you can switch back and forth between them infinite times). 
I could wrap the second arg in an object like `(visType, options)` and pass in `replacePeriodItems(myVisType, { useAltData })` if that makes it more clear than just passing the string as the second arg though?

jenniferarnesen (approved):

PR #1490, opened 2020-12-03
feat: scatter context menu + data tabs (DHIS2-10073)

martinkrulltott:
Implements [DHIS2-10073](https://jira.dhis2.org/browse/DHIS2-10073)

---

### Key features

1. Add "Vertical" and "Horizontal" tabs to the Data dimension selector
2. Add context menus for vertical and horizontal in layout with a _Swap_ option
3. Fix the broken Cypress tests caused by Scatter
4. Add new Cypress tests for the new Scatter features
5. Implement and test saving Scatter charts ([DHIS2-10005](https://jira.dhis2.org/browse/DHIS2-10005))
6. Bumps `@dhis2/cypress-commands` and `@dhis2/cli-utils-cypress` to v4 which in turn bumps `Cypress` to v6

---

### Description

Continues the work on Scatter where https://github.com/dhis2/data-visualizer-app/pull/1472 left off.

Adds a new tab bar to the Data dimension selector for switching between Vertical and Horizontal. When accessed from the vertical / horizontal data items in the layout the respective tab is preselected. When accessed from the Data item in the dimensions panel, the vertical tab is preselected.
Switching between the tabs switches the selected items in the data dimension selector to the corresponding data set.

Adds context menus to the layout items in horizontal and vertical which has a disabled assigned categories item and an item for swapping. Since this menu is different from the current `ChipMenu` I refactored `Chip` to pass in the `ContextMenu` as a prop. 

Saving now works and the vertical/horizontal items are correctly reduced to one item each upon save, similar to how Single value and Gauge only saves the first dx item.
Example scenarios:

- Vertical A, Horizontal B: Saved as Vertical A, Horizontal B
- Vertical A, B, Horizontal C, D: Saved as Vertical A, Horizontal C
- Vertical A, B, C, D, Horizontal E: Saved as Vertical A, Horizontal E
- Vertical A, B, Horizontal B: Saved as Vertical A, Horizontal B

Implements a new Cypress-specific integration file (`scatter.cy.js`) which runs through all scatter features so far (listed above).
- Layout validation errors for vertical and horizontal
- Locked state for vertical, horizontal, points
- Warning state for vertical, horizontal
- Swapping
- Accessing the new tabs in the data dimension selector
- Save and delete

---

### Screenshots

image

image


packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
```diff
@@ -303,14 +306,50 @@ export class DialogManager extends Component {
...
                         displayNameProp={displayNameProperty}
                         selectedDimensions={selectedItems}
                         infoBoxMessage={infoBoxMessage}
                         {...props}
                     />
                 )
+                const dataTabs = [
+                    ITEM_ATTRIBUTE_VERTICAL,
+                    ITEM_ATTRIBUTE_HORIZONTAL,
+                ].includes(dialogId) ? (
```
edoardo, line 320: This is adding both tabs regardless of which item attribute is matched, is that correct?

packages/app/src/components/Layout/ScatterLayout/ScatterAxis.js
```diff
@@ -1,60 +1,179 @@
...
+            DIMENSION_ID_DATA,
+            horizontalItems,
+            ITEM_ATTRIBUTE_HORIZONTAL
+        )
+        setItemAttributes(
+            DIMENSION_ID_DATA,
+            verticalItems,
+            ITEM_ATTRIBUTE_HORIZONTAL
+        )
+        setItemAttributes(
```
edoardo, line 63: Is this causing multiple re-renders in components where these item attributes are used?

edoardo (approved):
Overall looks good.

packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
martinkrulltott: Yeah this is adding both tabs (screenshot) to be displayed and the current tab (depending on which layout item was clicked) is then getting selected by `selected={key === dialogId}` further down.
image

packages/app/src/components/Layout/ScatterLayout/ScatterAxis.js
martinkrulltott: Good question, I'm not exactly sure how this works. I only "see" one update, but maybe there are actually 4 re-renders bundled as one?
Maybe @janhenrikoverland or @amcgee could weigh in? 
The option I guess is to create a new action e.g. `acSwapItemAttributes` which handles this in one go?

packages/app/src/components/DimensionsPanel/Dialogs/DialogManager.js
edoardo: 👍

cypress/elements/dimensionModal/dataDimension.js
```diff
@@ -21,13 +22,13 @@ export const removeAllDataItems = () => cy.getBySel(removeAllButtonEl).click()
...
         .getBySel(selectedItemEl)
-        .should('not.be.visible')
+        .should('be.visible')
         .and('have.length', 0)
 
 export const expectDataItemsAmountToBeSelected = amount =>
     cy
         .getBySel(selectedItemEl)
-        .should('not.be.visible')
+        .should('be.visible')
```
martinkrulltott, line 28: technically out of scope, but this fn is unused and I noticed a typo in the assertion that I thought I'd fix while at it

jenniferarnesen:
I think that the code is not supposed to call `registerCommands` any more?  (in `cypress/support/index.js`). Also, have you checked into the `enableAutoLogin`? Not sure if it makes sense for DV or not, but you might be able to remove the line: https://github.com/dhis2/data-visualizer-app/blob/master/cypress/support/server.js#L11

Actually, what I would consider doing is to make a separate branch/PR for the cypress upgrade, to keep it separate from the Scatter chart-specific implementation.

PR #1511, opened 2020-12-14
test: cypress tests need to confirm relevant elements exist in DOM before getting redux state

jenniferarnesen:
Some tests were failing on CI but passing locally. Throttling the network connection enabled reproducing locally.

The issue was that the redux state was fetched prior to all the state changes completing. By ensuring the elements to be tested are present in the DOM before getting the redux state, we know that the redux state is current. (read about it here: https://www.cypress.io/blog/2018/11/14/testing-redux-store/)

I also found that sometimes checking that the visualization container was visible wasn't enough and that actually checking for the chart title element was necessary. Not sure why since the whole svg should appear at once, I would think.

Note that a few cypress tests are still failing due to what I think could be a backend issue (also present in 2.35.1). But if you compare tests on this branch to some other branches with the cypress dashboard, you'll see that most of the tests are fixed. https://dashboard.cypress.io/projects/sojh88


edoardo (approved):
For what I can tell it looks fine.
@martinkrulltott should probably review this too.

jenniferarnesen:
> For what I can tell it looks fine.
> @martinkrulltott should probably review this too.

I'll go ahead and merge it for now, and walk through it with him when he's back.

PR #1552, opened 2020-12-28
feat: scatter options (DHIS2-10006)

martinkrulltott:
Implements [DHIS2-10006](https://jira.dhis2.org/browse/DHIS2-10006)
Design spec [DHIS2-9778](https://jira.dhis2.org/browse/DHIS2-9778)

**Requires https://github.com/dhis2/analytics/pull/751**

---

### Key features

1. Implements all axis options for Scatter
2. Moves a bunch of options to the new `axes` and `legend` format from [DHIS2-10054](https://jira.dhis2.org/browse/DHIS2-10054).
3. All options components in `VisualizationOptions/Options/` that were previously tied to an axis (e.g. `CategoryAxisLabels`) have now been refactored to use an `axisId` prop to be generic and axis agnostic.
4. Adds a `NumberBaseType` that specifically handles text inputs with numbers

---

### Description

Scatter axis options differ from regular axis options as both the vertical and horizontal axes are range axes, i.e. the user can set the range, decimals and steps, as was previously only possible on the vertical axis. Regression lines are also available on both axes, which isn't possible on any other vis type.

To support this a new backend format is used, which is based on `axes` and `legend` format ([DHIS2-10054](https://jira.dhis2.org/browse/DHIS2-10054)). The following options were removed in favour for the new format:
`baseLineLabel`, `baseLineValue`, `domainAxisLabel`, `hideLegend`, `rangeAxisDecimals`, `rangeAxisLabel`, `rangeAxisMaxValue`, `rangeAxisMinValue`, `rangeAxisSteps`, `targetLineLabel`, `targetLineValue`.

The new format is implemented through a new action in the store, `acSetUiOption`, which uses option consts (e.g. `OPTION_AXIS_STEPS`) to switch on to put the requested option in the right place in the format. I.e. the store replicates the backend structure for `axes` and `legend`. The `SET_UI_OPTION` case in `reducers/ui.js` is rather long and should probably be extracted to a separate file, I'm open to suggestions on how to do this and when (either now or in a separate refactor PR later).

As the new `axes` object is identified by both `index` and `type` (e.g. `index: 0, type: 'RANGE'`, a new concept of `axisId` was introduced which concatenates the two props in to one (e.g. `axisId: RANGE_0`), to make for easier handling and the need to only pass a single prop around. The name clashes with another use of `axisId` (however they're not yet used in the same context), so renaming this might be a good idea if needed in the future. 

To facilitate the fact that regression lines can now be both vertical and horizontal at the same time (previously they could only be either, most where horizontal but Bar and Stacked Bar have vertical lines since they're vertical types), the text align options had to be refactored to cater for this. Instead of passing the `visType` as `getTextAlignOptions(fontStyleKey, visType)`, the `isVertical` prop is passed instead and the caller is responsible for calculating its value. 
_Gotcha_: Note that the text align options are used in two different contexts, 1) for adapting the labels used in the options dialog, 2) for the generating the correct Highcharts config values. These are easy to mix up at first glance.

To reduce code duplication the concepts of `BaseLine` and `TargetLine` have been replaced by the generic `RegressionLine`, with corresponding `RegressionLineValue` and `RegressionLineTitle`. Analytics has also replaced `FONT_STYLE_TARGET_LINE_LABEL` and `FONT_STYLE_BASE_LINE_LABEL` with `FONT_STYLE_REGRESSION_LINE_LABEL`.

---

### TODO

-   [x] Add reducer tests

---

### Known issues

-   [ ] This PR primarily focuses on the Axis tab in the Options modal, so the rest of the tabs will have to be revisited in a future PR.
-   [x] Regression line with empty string label is unchecked after saving, since backend doesn't return empty values. Use an "enabled" prop instead.
-   [ ] The vertical tab bar on the Axis tab isn't implemented yet: https://jira.dhis2.org/browse/DHIS2-9878

---

### Screenshots

_Data tab for Scatter_
image

_Axes tab for Scatter_
image

_Scatter chart with custom font styles, axis options and regression lines_
image

_Example of the new store format for `axis` and `legend` in use_
image


janhenrikoverland:
Add reducer tests

janhenrikoverland:
Fix impure `enabled` assignment on `axes`

packages/app/src/modules/options.js
```diff
@@ -200,26 +155,32 @@ export const getOptionsForRequest = () => {
...
     }
 
-    optionsFromVisualization.baseLine =
-        isNotDefault(optionsFromVisualization, 'baseLineLabel') ||
-        isNotDefault(optionsFromVisualization, 'baseLineValue')
-
-    optionsFromVisualization.targetLine =
-        isNotDefault(optionsFromVisualization, 'targetLineLabel') ||
-        isNotDefault(optionsFromVisualization, 'targetLineValue')
+    optionsFromVisualization.axes = optionsFromVisualization.axes?.map(axis => {
```
edoardo, line 164: Perhaps the `?` is redundant as we discussed.
(Also it was the first time I saw it used to check the existence of a function of a primitive type, so I wasn't sure its use in this context is legit)
```suggestion
    optionsFromVisualization.axes = optionsFromVisualization.axes.map(axis => {
```

packages/app/src/modules/options/scatterConfig.js
```diff
@@ -0,0 +1,90 @@
...
+                <AxisDecimals axisId={horisontalAxisId} />,
+                <AxisLabels axisId={horisontalAxisId} />,
+                <TargetLine axisId={horisontalAxisId} isVertical={true} />,
+                <BaseLine axisId={horisontalAxisId} isVertical={true} />,
+            ]),
+        },
+    ]),
+    getSeriesTab(),
+    getStyleTab([getTitlesSection()]),
+    getOutliersTab([
```
edoardo, line 84: Does this resolve to an empty tab?
Should it be removed until specific options for outliers are available?

edoardo (approved):
Approving.
PIease have a look at the comments and change requests before merging.

packages/app/src/components/VisualizationOptions/InfoText.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { IconInfo16 } from '@dhis2/ui-icons'
```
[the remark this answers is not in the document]
martinkrulltott: 👍

packages/app/package.json
```diff
@@ -20,6 +20,7 @@
         "@dhis2/d2-ui-interpretations": "^7.1.3",
         "@dhis2/data-visualizer-plugin": "35.17.0",
         "@dhis2/ui": "^6.1.3",
+        "@dhis2/ui-icons": "^6.1.3",
```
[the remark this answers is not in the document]
martinkrulltott: 👍

packages/app/src/modules/options/scatterConfig.js
martinkrulltott: Yes, I added an empty tab in hope of outliers being part of this PR, but as the PR grew too large and outliers aren't ready yet it'll come in a later PR instead. As noted in the **Known issues** in the description above, 

> This PR primarily focuses on the Axis tab in the Options modal, so the rest of the tabs will have to be revisited in a future PR.

PR #1559, opened 2021-01-12
fix: limit values (DHIS2-10249) and skip rounding (DHIS2-10251)

martinkrulltott:
Implements [DHIS2-10249](https://jira.dhis2.org/browse/DHIS2-10249) and [DHIS2-10251](https://jira.dhis2.org/browse/DHIS2-10251)

---

### Key features

1. Limit values added to most vis types
2. Limit values prevented from bleeding in to unsupported vis types
3. Skip rounding added to all vis types

---

### Description

Limit values was previously only available for PT, but the option would bleed (i.e. unexpectedly affect) in to other vis types. This option is now available in the options modal for all vis types except Gauge, SV and Scatter. The bleeding in to these excepted vis types has been stopped by explicitly removing `options.measureCritieria` in the ui adapters.

Skip rounding was also only available for PT and SV and was also bleeding to other vis types. Skip rounding has now been added to all vis types. 

---

### Screenshots

_Skip rounding option available for a column chart_
image

_the results of using Skip rounding_
image

_Limit values tab and options available for a column chart_
image

_the results of using Limit values_
image


edoardo (approved):
Looks good.

One note tho, should the text `This will apply to the entire table` be changed since now the `MeasureCriteria` option is used in other vis types than PT?

PR #1560, opened 2021-01-13
fix: YOY charts with relative week periods DHIS2-9729

edoardo:
Fixes [DHIS2-9729](https://jira.dhis2.org/browse/DHIS2-9279)

**Requires https://github.com/dhis2/analytics/pull/758**

---

### Description

Reworked the way data series for YOY charts which uses relative week periods are generated.
To avoid the problems described in the JIRA ticket, where data points are not correctly aligned when comparing week ranges of different years, the logic is now based on a matrix of xAxis index and period ids, which is generated from the analytics responses in the DV plugin.
From the matrix a map object is computed, and used to get the correct xAxis position where to plot a value based on the period id.
This solves both the problem of week 53 that some years have, and the problem of different week ranges for different years when using a relativePeriodDate.

Requesting the last 4 weeks from today this is what happens:

1. a metaData request is fired against the analytics api for the most recent year in the series (2021) passing *today* as relativePeriodDate
2. from the response, the last period is considered: W1
3. for all the other years in the series (2019 and 2020) a metaData request is made for getting information about W1 of each of those years
4. in the responses, the endDate of the week + 1 day is used as relativePeriodDate for each year (needed in the next step)
5. for each year in series, analytics data + metaData requests are made using the correct relativePeriodDate
6. the responses are used to build the matrix and the map, adding the "extra" week(s) where necessary and adding a null value for the years which don't have those weeks

---

### TODO

-   a similar issue happens when using relative day period and Feb 29 falls within the dates range

---

### Screenshots

This example is explained as follows:
- last 4 weeks for 2019 are: 2019W1, 2018W52, 2018W51, 2018W50 *no data for these*
- last 4 weeks for 2021 are: 2021W1 (no analytics data for 2021 yet), 2020W53, 2020W52, 2020W51
- last 4 weeks for 2020 are: 2020W1, 2019W52, 2019W51, 2019W50

The X axis gets then 5 labels, from W50 to W1 including W53
The line for 2020 has a gap corresponding to W53
The line for 2021 starts 1 week "later" compared to 2020, and has a value for week 53

Screenshot 2021-01-13 at 11 55 54

### DIFF
--- a/packages/plugin/src/api/analytics.js
+++ b/packages/plugin/src/api/analytics.js
@@ -2,10 +2,11 @@ import {
     Analytics,
     VIS_TYPE_PIVOT_TABLE,
     layoutGetDimensionItems,
-    getRelativePeriodsOptionsById,
-    WEEKS,
     DIMENSION_ID_PERIOD,
+    WEEKS,
 } from '@dhis2/analytics'
+
+import { getRelativePeriodTypeUsed } from '../modules/analytics'
 
 const peId = DIMENSION_ID_PERIOD
 
@@ -56,11 +57,7 @@ export const apiFetchAnalyticsForYearOverYear = async (
     console.log('pe items', peItems)
 
     // relative week period in use
-    if (
-        getRelativePeriodsOptionsById(WEEKS)
-            .getPeriods()
-            .find(p => p.id === peItems[0].id)
-    ) {
+    if (getRelativePeriodTypeUsed(peItems) === WEEKS) {
         console.log('relative weeks period in use')
         const relativeWeeksData = await prepareRequestsForRelativeWeeks({
             analyticsEngine,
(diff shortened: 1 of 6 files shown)


packages/plugin/src/api/analytics.js
```diff
@@ -38,28 +46,187 @@ export const apiFetchAnalyticsForYearOverYear = async (
...
     const currentMonth = ('' + (now.getMonth() + 1)).padStart(2, 0)
 
-    yearlySeriesRes.metaData.dimensions[peId].forEach(period => {
-        yearlySeriesLabels.push(yearlySeriesRes.metaData.items[period].name)
+    const peItems = layoutGetDimensionItems(visualization, peId)
+    console.log('pe items', peItems)
+
+    // relative week period in use
+    if (getRelativePeriodTypeUsed(peItems) === WEEKS) {
+        console.log('relative weeks period in use')
```
martinkrulltott, line 61: leftover log from dev?

edoardo: Yes, along many others.
I cleaned up all the debug logs.

martinkrulltott (approved):
LGTM. Obviously a bit tricky to follow all the details of the inner workings of the algoritm, but I think the solid addition of the test cases proves that this works. Great job 👍 

I suggest restructuring the test cases so each test is independent (instead of having its different parts being spread across different arrays).  This will make it easier to read and quicker to add new test cases as well.

PR #1579, opened 2021-01-21
fix: update analytics deps

janhenrikoverland:
Updates analytics deps to 4.5.29.


amcgee (approved):
LGTM - I think `4.x` and `^4` do the same thing though?

PR #1590, opened 2021-01-29
fix: feb 29 in YOY with daily relative period (DHIS2-9729)

edoardo:
Implements [DHIS2-9729](https://jira.dhis2.org/browse/DHIS2-9729) fix for Feb 29 edge case

**Requires https://github.com/dhis2/analytics/pull/758**

Note this branch is based on `fix/yoy-weekly`.

---

### Key features

1. Fix the edge case of February 29 when using relative days period in YOY.

---

### Description

The issue is similar to the W53 when using a relative weeks period, as not all years are leap years and depending on which years are selected in Series for a YOY chart, the February 29 "extra" day need to be handled, adding the value for the leap years and leaving a "hole" for non-leap years.

### Screenshots

The following examples show charts for 2020 (a leap year) and 2021 (a non-leap year).
If the range of days requested includes Feb 29, the charts should show 29-02 in the xAxis and if a value is available for that date on leap years, it should be plotted in the chart.
On dev there is currently no value for 2020-02-29, but the xAxis label is correctly positioned.
For 2021 the line should always start 1 day before the 2020 line.

"starting date": 2021-03-01:
Screenshot 2021-01-28 at 17 16 30

"starting date": 2021-03-14:
Screenshot 2021-01-28 at 17 14 38

"starting date": 2021-03-05:
Screenshot 2021-01-28 at 17 05 19


packages/plugin/src/api/analytics.js
```diff
@@ -106,6 +113,39 @@ export const apiFetchAnalyticsForYearOverYear = async (
     }))
 }
 
+const prepareRequestsForRelativeDays = async ({
```
jenniferarnesen, line 116: Is this an async function? At least I didn't see anything async about it.

edoardo: Good catch. Copy-paste mistake.

jenniferarnesen (approved):

PR #1591, opened 2021-01-29
fix: add a warning in PT Parameters tab (DHIS2-10269)

edoardo:
Implements [DHIS2-10269](https://jira.dhis2.org/browse/DHIS2-10269)

---

### Key features

1. Warn the user about PT options that only apply to legacy standard reports

---

### Screenshots
Screenshot 2021-01-29 at 15 37 36

### DIFF
--- a/packages/app/src/components/VisualizationOptions/NoticeBox.js
+++ b/packages/app/src/components/VisualizationOptions/NoticeBox.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { NoticeBox as UiNoticeBox } from '@dhis2/ui'
+
+import { tabSectionOption } from './styles/VisualizationOptions.style.js'
+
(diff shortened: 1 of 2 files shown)


packages/app/src/components/VisualizationOptions/NoticeBox.js
```diff
@@ -0,0 +1,29 @@
...
+        title,
+    }
+
+    if (type) {
+        props[type] = true
+    }
+
+    return (
+        <div className={tabSectionOption.className}>
+            <UiNoticeBox {...props}>{text}</UiNoticeBox>
```
martinkrulltott, line 18: A bit magical to pass the props like this, why not pass them explicitly?

edoardo: Refactored to use the same interface as `ui` NoticeBox.

martinkrulltott (approved):

PR #1603, opened 2021-02-09
fix: get the errorCode for the analytics response in the right place

edoardo:
Implements [DHIS2-10329](https://jira.dhis2.org/browse/DHIS2-10329)

---

### Key features

1. fix regression where the custom error screens disappeared

---

### Description

The previous code expected the `errorCode` at the root level of the response object.
Now it is available under `details`.

---

### Screenshots

Before:
Screenshot 2021-02-09 at 16 31 59

After:
Screenshot 2021-02-09 at 16 28 46

### DIFF
--- a/packages/app/src/components/Visualization/Visualization.js
+++ b/packages/app/src/components/Visualization/Visualization.js
@@ -42,7 +42,7 @@ export class Visualization extends Component {
     onError = response => {
         let error
         if (response) {
-            switch (response.details.errorCode) {
+            switch (response.details?.errorCode) {
                 case 'E7114':
                     error = new AssignedCategoriesDataElementsError()
                     break
--- a/packages/app/src/components/Visualization/Visualization.js
+++ b/packages/app/src/components/Visualization/Visualization.js
@@ -42,7 +42,7 @@ export class Visualization extends Component {
     onError = response => {
         let error
         if (response) {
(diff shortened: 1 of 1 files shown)


packages/app/src/components/Visualization/Visualization.js
```diff
@@ -42,7 +42,7 @@ export class Visualization extends Component {
     onError = response => {
         let error
         if (response) {
-            switch (response.errorCode) {
+            switch (response.details.errorCode) {
```
martinkrulltott, line 45:
```suggestion
            switch (response.details?.errorCode) {
```

martinkrulltott: can we always expect `details` or should we `?` it?

edoardo: My understanding looking at the dataEngine code is that the full api response always ends up in `details` for response statuses outside the 200-399 range.
See:
https://github.com/dhis2/app-runtime/blob/master/services/data/src/links/RestAPILink/fetchData.ts#L17
https://github.com/dhis2/app-runtime/blob/master/services/data/src/links/RestAPILink/fetchData.ts#L40

Adding the check does not harm tho.

martinkrulltott (approved):

PR #1624, opened 2021-02-25
fix: store metadata also when no data is returned

edoardo:
### Key features

1. fix bug which caused dimension ids instead of names to be shown in the layout chip tooltips

---

### Description

See attached screenshots.
The bug only happens for saved visualizations when no data is returned.
Metadata is always present in the response, but when there is no data to show, the `EmptyResponseError` was thrown before storing the metadata for the dimension selection.
The fix makes sure the metadata in the Redux store is always updated, and any error is thrown afterwards.

### Screenshots

Before:
Screenshot 2021-02-25 at 12 12 48

After:
Screenshot 2021-02-25 at 12 13 03

### DIFF
--- a/packages/app/src/components/Visualization/Visualization.js
+++ b/packages/app/src/components/Visualization/Visualization.js
@@ -74,11 +74,6 @@ export class Visualization extends Component {
 
     onResponsesReceived = responses => {
         const forMetadata = {}
-        if (
-            !responses.some(response => response.rows && response.rows.length)
-        ) {
-            throw new EmptyResponseError()
-        }
         responses.forEach(response => {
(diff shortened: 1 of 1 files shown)


martinkrulltott (approved):

martinkrulltott:
FYI this fix also solves the related buggy behaviour where dimension items are missing when a dimension without metadata is opened (e.g. the period 2017 is selected but is missing metadata, the period tooltip then says "2017" but when clicking on period no items are selected in the period dimension modal). Ticket [DHIS2-12130](https://jira.dhis2.org/browse/DHIS2-12130)

PR #1633, opened 2021-03-01
fix: minor tweaks for outlier options

martinkrulltott:
### Key features

1. Only display extreme lines % when extreme lines are enabled
2. Proper help text for detection methods
3. Add and improve Cypress tests for outliers and scatter

---

### Screenshots

_new help text for detection methods_

image

_extreme lines % only display when extreme lines are enabled_

image


cypress/integration/options/axes.cy.js
```diff
@@ -86,19 +88,60 @@ describe('Options - Vertical axis', () => {
...
+            enableAxisTitle(TEST_AXIS)
+        })
+        it('set custom title', () => {
+            setAxisTitle(TEST_AXIS, TEST_TITLE)
+        })
+        it('click the modal update button', () => {
+            clickOptionsModalUpdateButton()
+            expectChartTitleToBeVisible()
+        })
+        it(`config has vertical axis title "${TEST_TITLE}"`, () => {
```
jenniferarnesen, line 133: Just confirm that you are checking the `vertical` axis with the function `expectWindowConfigXAxis`

cypress/elements/optionsModal/outliers.js
```diff
@@ -0,0 +1,8 @@
+const outliersCheckboxEl = 'option-outliers-enabled-checkbox'
+
+export const enableOutliers = () =>
+    cy
+        .getBySel(outliersCheckboxEl)
+        .find('[type="checkbox"]')
+        .check({ force: true })
```
jenniferarnesen, line 7: Curious why `force: true` is needed.

jenniferarnesen (approved):

cypress/elements/optionsModal/outliers.js
martinkrulltott: I've added a comment to fix it here, https://github.com/dhis2/data-visualizer-app/blob/520be2b846f5e14c33e4bf923a0c258158d1def2/cypress/elements/optionsModal/verticalAxis.js#L10. I couldn't get the ui checkbox to checked, because the input field is hidden, so `.find('[type="checkbox"]')` finds a hidden element and refuses to check it. I could instead `click` on the `label`, but then I can't verify the `.should('be.checked')` on the line below. 
I could post a question about this to the UI Q&A though.

edoardo (approved):

cypress/elements/optionsModal/outliers.js
martinkrulltott: Just did some testing and this can be solved by adding a click first and then chaining the find afterwards, like so: 
```
cy.getBySel(outliersCheckboxEl)
    .click()
    .find('[type="checkbox"]')
    .should('be.checked')
```

cypress/integration/options/axes.cy.js
martinkrulltott: Good find, copy/paste error, the text should say `config has horizontal...`

cypress/elements/optionsModal/outliers.js
martinkrulltott: FYI, the reason for having the `should` in the first place was that I wanted it to wait for the checkbox to be checked before proceeding, since some checkboxes toggle other elements visibility. But when testing it looked like this worked even without the `should` assertion. However I'm keeping it anyway to avoid potential flakey tests in the future.

PR #1638, opened 2021-03-04
test: font style testing

martinkrulltott:
### Key features

1. Improve / extend tests for font style options
2. Minor refactor of common functions to reduce duplication (`expectAppToNotBeLoading`, `clickCheckbox`, `typeInput`, `clearInput`)
3. Minor improvement of Scatter by selecting more org units
4. Comment out a test in start.cy.js caused by a known backend bug

---

### Description

Tip: The vast majority of the changes are in `fontStyles.cy.js`, however don't compare the change, just look at the new file as a whole, as it's been completely refactored and GH tries to "compare" it in a weird way which just adds confusion.

### DIFF
--- a/cypress/elements/common.js
+++ b/cypress/elements/common.js
@@ -13,6 +13,14 @@ const loadingEl = 'dhis2-uicore-circularloader'
 export const expectAppToNotBeLoading = () =>
     cy.getBySel(loadingEl).should('not.exist')
 
+export const clickCheckbox = target =>
+    cy.getBySel(target).click().find('[type="checkbox"]').should('be.checked')
+
+export const typeInput = (target, text) =>
+    cy.getBySel(target).find('input').type(text)
+
+export const clearInput = target => cy.getBySel(target).find('input').clear()
+
 export const replacePeriodItems = (
     visType,
     options = { useAltData: false }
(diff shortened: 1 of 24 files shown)


jenniferarnesen:
Looks good. I am curious about the organization/name of the test files. If/when tests are added to test custom title/subtitles and color sets, will those go in a separate file (since "fontStyle" is testing font styles and not necessarily everything in the "style" tab)?

jenniferarnesen (approved):

martinkrulltott:
> Looks good. I am curious about the organization/name of the test files. If/when tests are added to test custom title/subtitles and color sets, will those go in a separate file (since "fontStyle" is testing font styles and not necessarily everything in the "style" tab)?

Thanks for the review! Yes the file structure is a good question. Since this is evolving organically its hard to say how this will end up or where its going, but I think testing on a feature basis is better than testing on the basis of the DOM structure (like e.g. testing all content of a tab). Both because the tab content can change over time (features being moved to different tabs) and because tabs differ very much between vis types.
So the goal of this PR was to test font styling as a feature, regardless of which tab each option belongs to. Some parts are TBC though, as listed here: https://github.com/dhis2/data-visualizer-app/pull/1638/files#diff-c8701120fc77d9e3ddc5a931c06d2e2ac87590256a5683fdd008ac9f5ae29968R501

PR #1667, opened 2021-03-22
fix: detect style changes from dashboard for re-rendering visualization

jenniferarnesen:
### Key features

1. Visualizations in dashboard were not rerendering to fit the item size in the following situations:
 * item fullscreen
 * edit item resize
 * change screen width (e.g., switch to/from small screen)
---

### Description
Dashboard and Data Visualizer apps have different approaches regarding visualization size and resizing. DV vis containers have display flex and use a counter that updates when the window is resized, to trigger a re-render of the visualization. In Dashboard, the container has a set width and height, which change on any of the above resizing scenarios. It does not use a counter.

The new effect will end up being specific to dashboard, while the existing effect that responds to `renderCounter` will be specific to DV. A future improvement would possibly be to use the same mechanisms for both apps. But playing it safe for now, since we are in a hard freeze.

### Screenshots

(the video clips off the legend at the bottom, but it is actually visible)

https://user-images.githubusercontent.com/6113918/111984635-e9e60880-8b0b-11eb-836e-bdc59abe4c7c.mov

### DIFF
--- a/packages/plugin/src/ChartPlugin.js
+++ b/packages/plugin/src/ChartPlugin.js
@@ -13,6 +13,7 @@ const ChartPlugin = ({
     animation: defaultAnimation,
 }) => {
     const canvasRef = useRef(undefined)
+    const prevStyle = useRef(style)
 
     const renderVisualization = useCallback(
         animation => {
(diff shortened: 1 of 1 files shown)


edoardo (approved):
I suppose for DV the style effect doesn't matter as the renderCounter one is used.

jenniferarnesen:
> I suppose for DV the style effect doesn't matter as the renderCounter one is used.

@edoardo I switched back to separate effects. Realized the problem was that I was using `prevStyle.width` instead of `prevStyle.current.width`. Oops...

PR #1691, opened 2021-04-12
refactor: use d2 adapter

edoardo:
Requires https://github.com/dhis2/app-runtime-adapter-d2/pull/85

### Key features

1. replace the local D2Shim implementation with the d2 adapter
2. implement a context provider for user settings
3. pass the locale from user settings to the D2Shim to avoid an additional request to user settings

---

### Description

Similarly to dashboard app, use the d2 adapter in place of the local D2Shim implementation.
While working on this I noticed multiple requests (mostly identical) to `userSettings` api are fired.
One of those is in the d2 adapter itself, needed to get the user locale used to initialise i18n.
Since DV also does a request to userSettings, we can pass the locale to the shim from the app, and skip the userSettings request in the shim.


yarn.lock
```diff
@@ -1526,6 +1526,14 @@
   dependencies:
     moment "^2.24.0"
 
+"@dhis2/app-runtime-adapter-d2@^1.0.2":
+  version "1.0.2"
+  resolved "https://registry.yarnpkg.com/@dhis2/app-runtime-adapter-d2/-/app-runtime-adapter-d2-1.0.2.tgz#b1a88d371830bfe63fbc00068f7eda98f3c9259f"
+  integrity sha512-HbkkRdt4qMHAv3EKjZUuXXm7QfHK7NuxEHhxR33xxLfZHGdI+doRqvqzmpvJYkPDv4PaY91V4FKgnyznbDdlkQ==
+  dependencies:
+    "@dhis2/ui-core" "^4.6.1"
```
martinkrulltott, line 1534: Any way we can avoid having two versions of `@dhis2/ui-core`?

edoardo: Yes, the `@dhis2/app-runtime-adapter-d2` dependency had `@dhis2/ui-core` listed as dependency but was not used.
Now `@dhis2/app-runtime-adapter-d2` is updated.

packages/app/src/components/UserSettingsProvider.js
```diff
@@ -0,0 +1,49 @@
...
+
+    useEffect(() => {
+        async function fetchData() {
+            const { userSettings } = await engine.query({
+                userSettings: userSettingsQuery,
+            })
+
+            setSettings({
+                ...userSettings,
+                displayProperty: userSettings.keyAnalysisDisplayProperty,
```
jenniferarnesen, line 26: Is this property used by the app?

packages/app/src/AppWrapper.js
```diff
@@ -42,15 +48,46 @@ const AppWrapper = () => {
...
+                                {({ d2 }) => {
+                                    if (!d2) {
+                                        // TODO: Handle errors in d2 initialization
+                                        return null
+                                    } else {
+                                        return (
+                                            <App
+                                                d2={d2}
+                                                location={history.location}
+                                                baseUrl={baseUrl}
```
jenniferarnesen, line 75: Could the App component just use the `useConfig` hook directly or is it not a functional component?

packages/app/src/components/UserSettingsProvider.js
jenniferarnesen: Looks like this and the `extractUserSettings` function need to be tweaked to not repeat each other.

packages/app/src/components/UserSettingsProvider.js
```diff
@@ -0,0 +1,49 @@
+import React, { useContext, useState, useEffect, createContext } from 'react'
+import PropTypes from 'prop-types'
+import { useDataEngine } from '@dhis2/app-runtime'
+
+export const userSettingsQuery = {
+    resource: 'userSettings',
+    params: {
+        key: ['keyDbLocale', 'keyUiLocale', 'keyAnalysisDisplayProperty'],
```
jenniferarnesen, line 8: Is `keyDbLocale` used by the app? based on what is in `extractUserSettings` it doesn't seem so.

jenniferarnesen (approved):

packages/app/src/components/UserSettingsProvider.js
edoardo: Removed.

PR #1709, opened 2021-05-03
fix: start screen vis type crash

martinkrulltott:
### Description

The "most viewed" section works by first fetching the "favourites" and then fetching the types of each item.
If, for whatever reason, there's a mismatch between these two results (i.e. not all favourites have a corresponding item in the type response) the app crashes.
This change prevents this from happening by excluding the mismatched items.
Will lead to favourites being missing from the list (maybe not ideal) but at least the app won't crash.

---

### Screenshots

_no visual changes_
image

### DIFF
--- a/packages/app/src/components/Visualization/StartScreen.js
+++ b/packages/app/src/components/Visualization/StartScreen.js
@@ -13,6 +13,7 @@ import history from '../../modules/history'
 import { VisualizationError, genericErrorTitle } from '../../modules/error'
 import { GenericError } from '../../assets/ErrorIcons'
 import { apiFetchVisualizations } from '../../api/visualization'
+import { matchVisualizationWithType } from './utils'
 
 const StartScreen = ({ error }) => {
     const [mostViewedVisualizations, setMostViewedVisualizations] = useState([])
@@ -32,19 +33,10 @@ const StartScreen = ({ error }) => {
                 )
                 const visualizationsWithType =
                     visualizationsResult.visualization.visualizations // {id: string, type: string}
-                const result = []
-
-                visualizations.forEach(visualization => {
-                    const type = visualizationsWithType.find(
-                        visWithType => visWithType.id === visualization.id
-                    )?.type
-                    if (type) {
-                        result.push({
-                            ...visualization,
-                            type,
-                        })
-                    }
-                })
+                const result = matchVisualizationWithType(
+                    visualizations,
+                    visualizationsWithType
+                )
 
                 setMostViewedVisualizations(result)
             }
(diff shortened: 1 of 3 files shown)


packages/app/src/components/Visualization/StartScreen.js
```diff
@@ -24,20 +24,27 @@ const StartScreen = ({ error }) => {
...
-                const result = visualizations.map(vis => ({
-                    ...visualizationsWithType.find(
-                        visWithType => visWithType.id === vis.id && visWithType
-                    ),
-                    ...vis,
-                }))
+                    visualizationsResult.visualization.visualizations // {id: string, type: string}
+                const result = []
+
+                visualizations.forEach(visualization => {
```
jenniferarnesen, line 37: Would it be possible to extract this to a function and add unit tests for it?

packages/app/src/components/Visualization/StartScreen.js
```diff
@@ -24,20 +24,27 @@ const StartScreen = ({ error }) => {
                 engine,
                 6
             )
-            const visualizations = mostViewedVisualizationsResult.visualization
+            const visualizations = mostViewedVisualizationsResult.visualization // {position: int, views: int, id: string, created: string}
```
jenniferarnesen, line 27: Remove comments on lines 27 and 34

martinkrulltott: I thought the comments were actually quite useful for future reference, to know the signature of the response. We don't do this anywhere else though, so I can remove them if you prefer.
Could prefix them with `Response: ` for clarification if that's better though.

packages/app/src/components/Visualization/StartScreen.js
martinkrulltott: Sure!

jenniferarnesen: Also, I'm wondering if it would be better to turn this around. By only fetching the ids from the `mostViewed` endpoint, you can do this:
```
const mostViewedVisualizationsResult = await apiFetchMostViewedVisualizations(
                engine,
                6
            )
            const mostViewedVisIds =
                mostViewedVisualizationsResult.visualization
            if (mostViewedVisIds && mostViewedVisIds.length) {
                const visualizationsResult = await apiFetchVisualizations(
                    engine,
                    mostViewedVisIds
                )
                const visualizationsWithType =
                    visualizationsResult.visualization.visualizations

                setMostViewedVisualizations(
                    visualizationsWithType.filter(({ id }) =>
                        mostViewedVisIds.includes(id)
                    )
                )
            }
```

jenniferarnesen: p.s., I didn't test it yet.

martinkrulltott: I think that would work. But what would the benefit be?

jenniferarnesen (approved):

PR #1716, opened 2021-05-07
fix: await ou levels

martinkrulltott:
### Key features

1. properly await `ouLevels` to prevent race condition

---

### Description

There's an issue on dashboards where dashboard items fail to load.
It's caused by [ouLevelUtils in Analytics](https://github.com/dhis2/analytics/blob/37b3c6650924ee92c948dd5eeaee12a03b1be159/src/modules/ouLevelUtils/index.js#L19) receiving undefined `ouLevels`, which are provided by https://github.com/dhis2/data-visualizer-app/blob/06f0426626d47c0f181cbf261ed6fc73c400a96a/packages/plugin/src/VisualizationPlugin.js#L92

As the `ouLevels` are fetched in a `useEffect`, there's a race condition where the fetching sometimes doesn't finish before the component returns. This is prevented by only returning if `ouLevels` are finished fetching.

---

### Screenshots

_error in dashboards_
image

### DIFF
--- a/packages/plugin/src/VisualizationPlugin.js
+++ b/packages/plugin/src/VisualizationPlugin.js
@@ -153,7 +153,7 @@ export const VisualizationPlugin = ({
         /* eslint-disable-next-line react-hooks/exhaustive-deps */
     }, [visualization, filters, forDashboard])
 
-    if (!fetchResult) {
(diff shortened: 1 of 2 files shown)


amcgee (approved):
LGTM

PR #1721, opened 2021-05-25
feat: hide periods based on system settings (DHIS2-11161)

martinkrulltott:
Implements [DHIS2-11161](https://jira.dhis2.org/browse/DHIS2-11161)

**Requires https://github.com/dhis2/analytics/pull/920**

---

### Key features

1. Populates `excludedPeriodTypes` with values from system settings
2. Changes all instances of `DAYS` and `WEEKS` from Analytics to `DAILY` and `WEEKLY`
1. Adds Cypress tests for the period dimension
1. Adds a new `containsExact` command in Cypress

---

### Description

_text_

---

### TODO

-   [ ] _task_

---

### Known issues

-   [ ] _issue_

---

### Screenshots

_supporting text_


cypress/integration/dimensions/period.cy.js
```diff
@@ -0,0 +1,484 @@
...
+        })
+        it('a period can be selected by button', () => {
+            selectItemByButton(defaultRelativePeriod)
+            expectItemToBeSelected(defaultRelativePeriod)
+        })
+        it('all can be unselected by button', () => {
+            unselectAllItemsByButton()
+        })
+    })
+    describe('relatives period', () => {
```
jenniferarnesen, line 98: "relative periods"

jenniferarnesen, line 98: Wondering if this describe and the next one on line 124 should be a subset of the `initial state` describe above since the test is not re-visiting the start page? Kind of encapsulates the default state where there are no system settings restrictions.

cypress/integration/dimensions/period.cy.js
```diff
@@ -0,0 +1,484 @@
...
+        })
+        it("period type is 'Quarters'", () => {
+            expectRelativePeriodTypeToBe('Quarters')
+        })
+        it("relative option 'Months' is not shown", () => {
+            openRelativePeriodsTypeSelect()
+            expectRelativePeriodTypeSelectToNotContain('Months')
+        })
+        it('all other relative options are shown', () => {
+            const relativePeriodTypes = [
```
jenniferarnesen, line 197: I guess this is where you talking about refactoring? Yeah, I agree there is a lot of repetition. What about having a master array `allRelativePeriods` and `allFixedPeriods`, and then doing something like: 
```
const relativePeriodsWithoutMonthly = allRelativePeriods.filter(p => p !== 'Monthly')
```

Then I guess you would have to create the object for the amountOfChildren tests from the `allRelativePeriods` array.

edoardo (approved):

cypress/integration/dimensions/period.cy.js
martinkrulltott: Yeah I considered it, but I think the tests are already too dynamic as they are. Best to keep it simple and explicitly provide the expect result so we can be sure that the test works. 
However I did change a lot of the naming of props and variables to be easier to read and follow. Hopefully that solves a bit of the confusion at least 🤞

cypress/integration/dimensions/period.cy.js
martinkrulltott: Good idea, done!

martinkrulltott: Oopsi 👍

jenniferarnesen (approved):

PR #1735, opened 2021-06-03
fix: drill down bugs (DHIS2-11061)

martinkrulltott:
Implements a fix for [DHIS2-11061](https://jira.dhis2.org/browse/DHIS2-11061)

**Requires https://github.com/dhis2/analytics/pull/931**

---

### Key features

1. Change the popup placement to `right-start`
2. Hide drill down from dual category when org unit is used as the first category
3. Prevent the popup from opening empty while its still loading.

---

### Description

1. Position `right` didn't work for tall items (columns), as it would put the popup in the vertical center. With `right-start` it goes to the top of the right side.
2. Drill down for dual category when the org unit is the first category doesn't make sense, as the column clicked doesn't represent an org unit. To prevent this, dual category drill down is only available when the column is an org unit (i.e. org unit is the second category)
3. Previously the popup would open directly, while it was still loading, to show just a sliver of the popup (see screenshot below). Now it only opens when it's finished loading.

---

### Screenshots

**BEFORE**
_popup opening in an empty state while loading_
image

_the content of the popup was only shown after the loading was finished_
image

**AFTER**

_`right-start` placement for dual category column_
image

_`right-start` placement for dual category stacked bar_
image

_`right-start` placement for PT_
image

_no popup available when the org unit is the first category_
image

### DIFF
--- a/packages/plugin/src/ContextualMenu.js
+++ b/packages/plugin/src/ContextualMenu.js
@@ -59,56 +59,54 @@ export const ContextualMenu = ({ config, ouLevels, onClick }) => {
         minWidth: 200,
     }
 
-    return (
+    return ouData ? (
         <FlyoutMenu>
-            {ouData && (
-                <MenuItem dense label={i18n.t('Change org unit')}>
-                    {ouData?.parent && (
-                        <>
-                            <MenuItem
-                                dense
-                                icon={<ArrowUpwardIcon />}
-                                label={
-                                    <span style={menuItemStyle}>
-                                        {ouData.parent.name}
-                                    </span>
-                                }
-                                onClick={() =>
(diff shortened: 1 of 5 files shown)


packages/plugin/src/ContextualMenu.js
```diff
@@ -59,56 +59,54 @@ export const ContextualMenu = ({ config, ouLevels, onClick }) => {
         minWidth: 200,
     }
 
-    return (
+    return ouData ? (
```
edoardo, line 62: This is fine for now.
I think originally it was different because we would have other drill options in the menu, like period.
This will need to change anyway when we implement period drilling.

edoardo (approved):

PR #1752, opened 2021-06-15
test: new tests for multi axis options (DHIS2-6672)

martinkrulltott:
Implements tests for [DHIS2-6672](https://jira.dhis2.org/browse/DHIS2-6672)

Tests the following options:

- Steps
- Decimals (can only be enabled but not evaluated)
- Auto generated axis titles for single and multi axis

image

### DIFF
--- a/cypress/elements/optionsModal/axes.js
+++ b/cypress/elements/optionsModal/axes.js
@@ -5,6 +5,8 @@ const rangeMinInputEl = 'axis-range-min-input'
 const rangeMaxInputEl = 'axis-range-max-input'
 const axisTitleRadiosEl = 'option-chart-title'
 const verticalTabsEl = 'axes-tabs'
+const stepsInputEl = 'axis-steps'
+const decimalsInputEl = 'axis-decimals'
 
 const getAxisSelector = (axis, selector) => `${axis}-${selector}`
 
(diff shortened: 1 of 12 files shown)


cypress/elements/optionsModal/axes.js
```diff
@@ -17,6 +19,12 @@ export const setAxisTitleText = (axis, text) =>
 export const setAxisTitleToCustom = () =>
     cy.getBySel(axisTitleRadiosEl).contains('Custom').click()
 
+export const setAxisTitleToAuto = () =>
```
jenniferarnesen, line 22: These 3 functions could be made into a single function `setAxisTitleTo(option)` or maybe `setAxisTitleOptionTo(option)`. In the test it would look like:

`setAxisTitleOptionTo('None')`

which one could argue is more readable than

`setAxisTitleOptionToNone()`

Just a thought. It's your call.

martinkrulltott: Good idea 👍

cypress/integration/options/axes.cy.js
```diff
@@ -152,3 +169,114 @@ describe('Options - Horizontal axis', () => {
...
+describe('Options - Auto-generated axis title', () => {
+    it('navigates to the start page and add a data item', () => {
+        goToStartPage()
+        openDimension(DIMENSION_ID_DATA)
+        selectDataElements([TEST_DATA_ELEMENT_NAME])
+        clickDimensionModalUpdateButton()
+        expectVisualizationToBeVisible(VIS_TYPE_COLUMN)
+    })
+    describe('Single item - single axis', () => {
+        it('opens Options -> Axes', () => {
```
jenniferarnesen, line 182: There is some inconsistency of the phrasing of the test names in the test files: e.g. open vs opens, set vs sets. See line 182 vs 186, 190.

Not a big deal, but it would read better if it were consistent.

jenniferarnesen (approved):

cypress/integration/options/axes.cy.js
martinkrulltott: Didn't see this comment until after I merged it (got trolled by GH... again). But I'll keep this in mind for sure!

PR #1793, opened 2021-07-07
fix: cypress test user has different permissions and access

jenniferarnesen:
Recent switch to `analyticstest` user for cypress tests. This user has somewhat different access than the previous user `admin`

Three changes:
- update number of items expected based on new user permissions
- enable the save test for scatter types
- added a check for loading for an intermittently failing test (not sure if this is a flaky solution though)


martinkrulltott (approved):

cypress/integration/save.cy.js
```diff
@@ -63,16 +66,16 @@ describe('saving an AO', () => {
...
         })
         it('adds Data dimension items', () => {
+            openDimension(DIMENSION_ID_DATA)
             if (TEST_VIS_TYPE === VIS_TYPE_SCATTER) {
-                cy.contains(
-                    'Scatter was randomly picked but this has not been implemented for Scatter yet, please rerun the test'
-                ).should('exist')
+                selectIndicators([TEST_INDICATOR_NAMES[0]])
+                switchDataTab('Horizontal')
+                selectIndicators([TEST_INDICATOR_NAMES[1]])
```
martinkrulltott, line 74: This fails (https://github.com/dhis2/data-visualizer-app/pull/1793/checks?check_run_id=3016234639) because if you try to select indicators twice, Cypress wants to click an option in a dropdown that's already selected, which ui doesn't support (raised a ticket for this ages ago). So you can either 

- select first an indicator, switch tab then a data element, 
- or close the dimension between the selections (open, select indicator, click update, open, switch tab, select indicator)
- or select an indicator, switch tab then select using `selectFirstDataItem`
- or select an indicator, switch tab, reset with `switchDataTypeToAll`, then select an indicator

martinkrulltott: https://jira.dhis2.org/browse/LIBS-74

PR #1797, opened 2021-07-08
feat: PT fixed headers (DHIS2-11057)

edoardo:
Implements [DHIS2-11057](https://jira.dhis2.org/browse/DHIS2-11057)

**Requires https://github.com/dhis2/analytics/pull/984**

---

### Key features

1. Add 2 options in the Style tab for PT: `fixColumnHeaders` and `fixRowHeaders`
2. Pass the options to the visualization plugin so they can be used by the PT engine/components

---

### Description

The fixed headers in PT consist of 2 separate features:
- fix row headers
- fix column headers

For these 2 separate options are needed, which can be used independently or combined together.

---

### Screenshots

PT options Style tab:
Screenshot 2021-08-19 at 12 39 35

### DIFF
--- a/packages/app/i18n/en.pot
+++ b/packages/app/i18n/en.pot
@@ -411,11 +411,11 @@ msgstr "Extreme line % detection"
 msgid "Number"
 msgstr "Number"
 
-msgid "Fixed column headers to top of table"
-msgstr "Fixed column headers to top of table"
+msgid "Fix column headers to top of table"
+msgstr "Fix column headers to top of table"
 
-msgid "Fixed row headers to left of table"
-msgstr "Fixed row headers to left of table"
+msgid "Fix row headers to left of table"
+msgstr "Fix row headers to left of table"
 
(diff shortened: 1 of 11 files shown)


packages/app/i18n/en.pot
```diff
@@ -402,6 +402,12 @@ msgstr "Extreme line % detection"
 msgid "Number"
 msgstr "Number"
 
+msgid "Fixed column headers to top of table"
```
martinkrulltott, line 405: Is this the final phrasing for these options, verified by @cooper-joe ?

cooper-joe: No, it should be `Fix column headers...` and `Fix row headers...`.

martinkrulltott (approved):

packages/app/i18n/en.pot
edoardo: Nice catch, fixed!

PR #1800, opened 2021-07-09
chore: upgrade react-redux to 7.2.4

jenniferarnesen:
Had to replace the old school `store.dispatch` in components, which is no longer available in the latest version of `react-redux`. 

The mock store in the tests still provide this so the tests are still using that, but we may want to review the jest tests and discard what we think isn't useful anymore, now that we have cypress.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -15047,7 +15047,7 @@ react-portal@^4.1.5:
   dependencies:
     prop-types "^15.5.8"
 
-react-redux@^5.0.7, react-redux@^5.1.0:
+react-redux@^5.0.7:
   version "5.1.2"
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -15047,7 +15047,7 @@ react-portal@^4.1.5:
   dependencies:
     prop-types "^15.5.8"
 
-react-redux@^5.0.7, react-redux@^5.1.0:
+react-redux@^5.0.7:
```
martinkrulltott, line 15050: why do we have both v5 and v7?

martinkrulltott (approved):

yarn.lock
jenniferarnesen: Looks like both react-beautiful-dnd and d2-ui-favorites-dialog use it

PR #1879, opened 2021-09-13
chore(deps): update app-runtime to latest

ismay:
https://jira.dhis2.org/browse/LIBS-200. Updating to the new app-runtime release.

Still todo: 

- [ ] Update analytics once this has been published: https://github.com/dhis2/analytics/pull/1000. Otherwise the versions of ui won't match

### DIFF
--- a/packages/plugin/src/__tests__/ChartPlugin.spec.js
+++ b/packages/plugin/src/__tests__/ChartPlugin.spec.js
@@ -34,13 +34,6 @@ const ouMock = {
     ],
 }
 
-const yearOverYearCurrentMock = {
-    type: analytics.VIS_TYPE_YEAR_OVER_YEAR_LINE,
-    columns: [dxMock],
-    rows: [peMock],
-    yearlySeries: ['LAST_YEAR'],
-}
-
 const mockExtraOptions = {
     dashboard: false,
     noData: {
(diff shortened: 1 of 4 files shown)


packages/plugin/src/__tests__/ChartPlugin.spec.js
```diff
@@ -161,15 +154,6 @@ describe('ChartPlugin', () => {
...
         })
 
-        describe('Year-on-year chart', () => {
-            beforeEach(() => {
-                props.visualization = {
-                    ...yearOverYearCurrentMock,
-                    option1: 'def',
-                }
-            })
-        })
```
ismay, line 171: I've removed this as there is no test in the describe block and it was failing with the new jest version.

martinkrulltott (approved):

martinkrulltott:
LGTM. Potentially dedupe yarn.lock with `npx yarn-deduplicate yarn.lock` before merging

ismay:
> LGTM. Potentially dedupe yarn.lock with `npx yarn-deduplicate yarn.lock` before merging

Yeah I'll do that. I'll update it with v8 of cli-app-scripts once that's out, and then if 2.37 has been branched it should be good to merge.

martinkrulltott:
Sounds good. 2.37 has been branched to [37.x](https://github.com/dhis2/data-visualizer-app/tree/37.x), so this can actually be merged whenever its ready.

PR #1914, opened 2021-10-01
fix: return a flat list of dimension items with multiple dimension on axis

edoardo:

### Key features

1. Bug fix

---

### Description

The `sGetDimensionItemsByAxis` incorrectly returned `undefined` if more than 1 dimension was selected on an axis.
The reason was that an array was passed to `sGetUiItemsByDimension` instead of a dimension id causing that function to return `undefined`.
The fix is looping through the dimensions array and call `sGetUiItemsByDimension` for each dimension before returning a flattened list of dimension items.

### DIFF
--- a/packages/app/src/reducers/__tests__/ui.spec.js
+++ b/packages/app/src/reducers/__tests__/ui.spec.js
@@ -686,7 +686,6 @@ describe('reducer: ui', () => {
                 },
             }
 
-            // TODO: what is the expected result?
             const expectedDimensions = ['anc', 'bcg', 'six-months']
 
             expect(ui.sGetDimensionItemsByAxis(state, AXIS_ID_ROWS)).toEqual(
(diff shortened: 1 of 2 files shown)


packages/app/src/reducers/__tests__/ui.spec.js
```diff
@@ -642,4 +642,56 @@ describe('reducer: ui', () => {
...
+                    },
+                    itemsByDimension: {
+                        [DIMENSION_ID_DATA]: ['anc', 'bcg'],
+                        [DIMENSION_ID_ORGUNIT]: ['tanzania'],
+                        [DIMENSION_ID_PERIOD]: ['six-months'],
+                    },
+                },
+            }
+
+            // TODO: what is the expected result?
```
jenniferarnesen, line 689: Remove the TODO comment, assuming the expected output is correct.

jenniferarnesen:
Seems to be a prettier issue in packages/app/src/reducers/ui.js

jenniferarnesen (approved):

packages/app/src/reducers/__tests__/ui.spec.js
edoardo: Done.

PR #1931, opened 2021-10-11
fix: displayNameProperty prop name typo + prevent duplicates in store + remove unused fn

martinkrulltott:
### Key features

1. Prevents `keyAnalysisDisplayProperty` and `keyUiLocale` from appearing twice in the store
2. Adds missing translations to `visTypeDescriptions`
3. Removes the unused fn `decodeDataTransfer`
4. Removes the unused file `storeDesign`
5. Removes the unused const `DATA_SETS_CONSTANTS` 
6. ~~Logs base url in Cypress~~

---

### Description
1. `keyAnalysisDisplayProperty` and `keyUiLocale` are renamed to `displayProperty` and `uiLocale` respectively, but the originals were still saved in the store too. This fix filters out the originals and only stores the renamed props.
2. Translations were missing because the i18n strings were wrapped in a const, not a function
3. Unused, so might as well remove it
4. Same as above
5. Same as above. Plus it wouldn't work with i18n anyway as its a const, not a function
6. ~~Logs the base url in Cypress for readability in the Cypress dashboard, to know which server is being used~~


martinkrulltott:
Waiting for https://github.com/dhis2/data-visualizer-app/pull/2125 to be merged first

edoardo (approved):

PR #1971, opened 2021-12-21
fix: canceled request when no data DHIS2-12130 v35

edoardo:
Fix for a bug related to [DHIS2-12130](https://jira.dhis2.org/browse/DHIS2-12130)

---

### Key features

1. fix canceled request for org unit levels

---

### Description

The fetch was not async, so when the analytics response was coming back first and there is no data available, the request to orgunit levels was canceled when the `VisualizationPlugin` component was replaced by the error screen for no data.
The issue does not happen from v36 and above.

---

### TODO

-   [ ] Fix failing tests

---

### Known issues

-   [ ] tests are failing

### DIFF
--- a/packages/plugin/src/VisualizationPlugin.js
+++ b/packages/plugin/src/VisualizationPlugin.js
@@ -2,12 +2,12 @@ import React, { useEffect, useState, useCallback } from 'react'
 import { createPortal } from 'react-dom'
 import PropTypes from 'prop-types'
 
-import { useDataEngine, useDataQuery } from '@dhis2/app-runtime'
+import { useDataEngine } from '@dhis2/app-runtime'
 import { Popper } from '@dhis2/ui'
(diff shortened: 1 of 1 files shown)


packages/plugin/src/VisualizationPlugin.js
```diff
@@ -30,6 +30,7 @@ export const VisualizationPlugin = ({
     ...props
 }) => {
     const engine = useDataEngine()
+    const [ouLevels, setOuLevels] = useState(undefined)
```
martinkrulltott, line 33: why explicitly set `undefined` instead of just `useState()`?

edoardo: I copied from v36 so no reason I guess 😁 
Fixed.

martinkrulltott (approved):

PR #1977, opened 2021-12-23
test: selecting level / group with user org unit

martinkrulltott:
Adapts the Cypress tests to the changes introduced in the latest Analytics version, specifically that when `User org unit` is selected (which is the new default state when creating a new AO), `Level` and `Group` dropdowns are disabled. ([Analytics PR](https://github.com/dhis2/analytics/pull/1100)).

Before accessing the `Level` or `Group` dropdowns, the `User org unit` needs to be deselected so that they are enabled again (plus making sure that `Sierra Leone` is selected).

### DIFF
--- a/cypress/integration/dimensions/orgUnit.cy.js
+++ b/cypress/integration/dimensions/orgUnit.cy.js
@@ -51,28 +51,6 @@ describe(`Org unit dimension`, () => {
         expectWindowConfigSeriesToHaveLength(1)
         expectDimensionToHaveItemAmount(DIMENSION_ID_ORGUNIT, 1)
     })
-    const TEST_LEVEL = 'District'
-    it(`selects a level - ${TEST_LEVEL}`, () => {
-        openDimension(DIMENSION_ID_ORGUNIT)
-        expectOrgUnitDimensionModalToBeVisible()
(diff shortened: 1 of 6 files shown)


cypress/integration/dimensions/orgUnit.cy.js
```diff
@@ -51,28 +51,6 @@ describe(`Org unit dimension`, () => {
...
-        expectOrgUnitDimensionToNotBeLoading()
-        toggleOrgUnitLevel(TEST_LEVEL)
-        clickDimensionModalUpdateButton()
-        expectVisualizationToBeVisible(VIS_TYPE_COLUMN)
-        expectWindowConfigSeriesToHaveLength(1)
-        expectDimensionToHaveItemAmount(DIMENSION_ID_ORGUNIT, 1)
-    })
     const TEST_DISTRICT_1 = 'Bo'
     it(`selects a district level org unit - ${TEST_DISTRICT_1}`, () => {
         openDimension(DIMENSION_ID_ORGUNIT)
```
martinkrulltott, line 56: Reordered, so that the Bo test comes first (it deselects User org units by running `deselectUserOrgUnit(TEST_DEFAULT_ORG_UNIT)`)

edoardo (approved):

PR #1985, opened 2022-01-13
fix: define DV vis types to use and descriptions in app

edoardo:
**Requires https://github.com/dhis2/analytics/pull/1126**

Using 22.0.0-alpha.2 published with the code from the above PR.

---

### Key features

1. fix the breaking change in analytics
2. define the vis types to use locally
3. move the vis type descriptions from analytics to the app

---

### Description

The above analytics PR has a breaking change for DV.
This PR accommodates for the new code in analytics.

The list of vis types supported by DV should not be a `analytics` library concern.
Also, the descriptions for the vis types handled by DV are only needed in DV so far.
The PT description for ER is different from the one for DV, so having them defined at app level simplifies the code.


cypress/utils/random.js
```diff
@@ -1,4 +1,38 @@
...
+    VIS_TYPE_PIE,
+    VIS_TYPE_RADAR,
+    VIS_TYPE_GAUGE,
+    VIS_TYPE_YEAR_OVER_YEAR_LINE,
+    VIS_TYPE_YEAR_OVER_YEAR_COLUMN,
+    VIS_TYPE_SINGLE_VALUE,
+    VIS_TYPE_SCATTER,
+} from '@dhis2/analytics'
+
+const visTypes = [
```
martinkrulltott, line 19: I think this should be placed somewhere else than `random.js`, as it needs to be used in other places as well (which are unrelated to the generation of random items). e.g. `open` and `new` uses the old list as the source for all vis types to be tested, those will need the `visTypes` list as well.

packages/app/src/components/MenuBar/MenuBar.js
```diff
@@ -1,11 +1,12 @@
-import { FileMenu } from '@dhis2/analytics'
+import { FileMenu, VIS_TYPE_ALL, VIS_TYPE_CHARTS } from '@dhis2/analytics'
```
martinkrulltott, line 1: Maybe `VIS_TYPE_ALL` and `VIS_TYPE_CHARTS` should be renamed, as they're not "types" themselves but groups of types, e.g. `VIS_TYPE_GROUP_ALL` ?

packages/app/src/modules/visualization.js
```diff
@@ -1,6 +1,90 @@
...
+    VIS_TYPE_SCATTER,
+} from '@dhis2/analytics'
+import i18n from '@dhis2/d2-i18n'
 import { DEFAULT_CURRENT } from '../reducers/current.js'
 import { DEFAULT_VISUALIZATION } from '../reducers/visualization.js'
-import options from './options.js'
+import { options } from './options.js'
+
+export const visTypes = [
+    { type: VIS_TYPE_PIVOT_TABLE },
```
martinkrulltott, line 24: why wrap each type in an object with a `type` prop instead of just keeping them directly in the array?
```suggestion
    VIS_TYPE_PIVOT_TABLE,
```

martinkrulltott (commented):
Looks good so far. I suggest releasing an alpha version of Analytics to make sure that all Cypress tests run properly and potential kinks can be worked out before the breaking changes is release.

packages/app/src/modules/visualization.js
edoardo: Because of this: https://github.com/dhis2/analytics/pull/1126/files#diff-71b1b0baf25e7ae70c0c640cb684eca4d5a689bf98576599ef003cdd8de0dd58R20

cypress/utils/random.js
edoardo: Moved the list to `utils/data.js` where the vis types from analytics are already imported.
Use this exported list also in `new.cy.js`.

packages/app/src/modules/visualization.js
martinkrulltott: Ok I see. From an app perspective, I think keeping it as simple as possible would be the most straight forward. We can always refactor this list internally in the future if needed (if this was a lib with an API it would be different though), so as long as the objects only contain a single prop the `type` prop just becomes an extra step.

packages/app/src/components/MenuBar/MenuBar.js
edoardo: Good point, renamed.

martinkrulltott (approved):

edoardo:
I started a separate branch where I somehow managed to solve the dependency issue. Closing this one.

PR #2017, opened 2022-04-04
fix: use translated name in title DHIS2-13015

edoardo:
Implements [DHIS2-13015](https://jira.dhis2.org/browse/DHIS2-13015)

**Requires https://github.com/dhis2/analytics/pull/XXX**

---

### Key features

1. use translated name in visualization title
2. perform the search on `displayName` instead of `name`

---

### Description

The title was using `visualization.name` which is the original name chosen when the visualization was saved.
Instead, `visualization.displayName` should be used, which contains the translated version in the current user language (when available).

The search in the `OpenFileDialog` was also performed on `name` instead of `displayName`.

---

### TODO

-   [ ] updated `analytics` dependency with the fix for the search in `OpenFileDialog`

---

### Known issues

-   [ ] blocked by https://dhis2.slack.com/archives/G3TL0J145/p1648477842131529

---

### Screenshots

Before:
Screenshot 2022-04-04 at 13 36 34

After:
Screenshot 2022-04-04 at 13 35 56

### DIFF
--- a/packages/app/src/components/TitleBar/TitleBar.js
+++ b/packages/app/src/components/TitleBar/TitleBar.js
@@ -35,7 +35,7 @@ const getTitleText = (titleState, visualization) => {
             return getTitleUnsaved()
         case STATE_SAVED:
         case STATE_DIRTY:
-            return visualization.name
+            return visualization.displayName || visualization.name
         default:
             return ''
     }


turban (approved):
LGTM!

PR #2103, opened 2022-06-08
feat: use About and Interpretations components from analytics

edoardo:
**Requires https://github.com/dhis2/analytics/pull/1250**

---

### Key features

1. fix the double Redux store issue
2. use AboutAOUnit from analytics
3. use InterpretationsUnit from analytics
4. use interpretation components from analytics (modal, unit...)
5. rewrite DownloadMenu component so it can be used in both FileMenu and InterpretationModal

---

### Description

_Coming soon_

---

### TODO

-   [x] update analytics dependency

---

### Known issues

-   [x] PT table does not show in the interpretation modal

---

### Screenshots

Right sidebar:
Screenshot 2022-06-13 at 09 25 16

Interpretation modal:
Screenshot 2022-06-13 at 09 25 36

Download menu in interpretation modal:
Screenshot 2022-06-13 at 09 27 05


packages/app/src/components/DownloadMenu/DownloadMenu.js
```diff
@@ -1,265 +1,55 @@
...
+            download={downloadData}
+            label={i18n.t('Microsoft Excel')}
+            format={FILE_FORMAT_XLS}
+        />
+        <PlainDataSourceSubMenu
+            download={downloadData}
+            label={i18n.t('CSV')}
+            format={FILE_FORMAT_CSV}
+        />
+        <AdvancedSubMenu download={downloadData} label={i18n.t('Advanced')} />
```
martinkrulltott, line 45: It's not entirely clear to me that `download`, `downloadData` and `downloadImage` are actually functions. 
For the submenus (that now use have a `download` prop), these might as well just be called `onClick`, but a name with a prefix that indicates that its actually a function (like "on...", "perform...", "execute...", "start...", "do...") would work as well I suppose.
For the `useDownloadMenu` props, it might be more clear that these are actually functions when looking at the context (however just `downloadData` can be misinterpreted as "the data returned from a download" instead of "perform a download of data"), so maybe they can be left as is, but preferably they'd also be prefixed in a way so its clear that these are actually functions.

packages/app/src/components/DownloadMenu/GraphicsMenu.js
```diff
@@ -0,0 +1,31 @@
...
+    MenuSectionHeader,
+    IconImage24,
+    IconFileDocument24,
+    colors,
+} from '@dhis2/ui'
+import PropTypes from 'prop-types'
+import React from 'react'
+import { FILE_FORMAT_PDF, FILE_FORMAT_PNG } from './constants.js'
+
+/* eslint-disable react/jsx-key */
```
martinkrulltott, line 13: why disable this rule instead of using a key?

packages/app/src/components/DownloadMenu/PlainDataSourceSubMenu.js
```diff
@@ -0,0 +1,57 @@
...
+    ID_SCHEME_UID,
+    ID_SCHEME_CODE,
+    ID_SCHEME_NAME,
+} from './constants.js'
+
+export const PlainDataSourceSubMenu = ({
+    download,
+    format,
+    label,
+    ...menuItemProps
```
martinkrulltott, line 16: `menuItemProps` is unused and should be removed?

packages/app/src/components/DownloadMenu/TableMenu.js
```diff
@@ -0,0 +1,41 @@
...
+    DOWNLOAD_TYPE_TABLE,
+    FILE_FORMAT_XLS,
+    FILE_FORMAT_CSV,
+    FILE_FORMAT_HTML_CSS,
+} from './constants.js'
+
+/* eslint-disable react/jsx-key */
+export const TableMenu = ({ download }) =>
+    React.Children.toArray([
+        <MenuSectionHeader label={i18n.t('Table layout')} hideDivider={true} />,
```
martinkrulltott, line 15:
```suggestion
        <MenuSectionHeader label={i18n.t('Table layout')} hideDivider />,
```

martinkrulltott (approved):

packages/app/src/components/DownloadMenu/PlainDataSourceSubMenu.js
edoardo: 🤔 it's passed to `<MenuItem>`.

martinkrulltott: What I mean is that it's never set in the first place (look at where this component is being used)

HendrikThePendric: Just to confirm. I agree with @martinkrulltott that these `menuItemProps` can and should be removed. They are currently unused, and it would probably make more sense to add explicit props if we do need to pass anything to the menu item later on.

packages/app/src/components/DownloadMenu/DownloadMenu.js
edoardo: Renamed props and functions.

PR #2125, opened 2022-07-21
fix: remove d2

edoardo:

---

### Key features

1. prepare to get rid of `d2` entirely

---

### Description

This branch has changes to remove the need for `d2`.
There is still one place where `d2` is needed and we don't have an alternative at the moment, the user data store.

---

### TODO

-   [x] figure out how to use the user data store without `d2`
-   [ ] lots of regression testing
-   [x] fix tests, particularly the App.js tests, but before touching them we should figure out how we want to test, what should 
be tested in Cypress and what in unit tests
-   [ ] look for and clean up more code not needed
-   [x] see if it's possible to avoid fetching all user authorities to check for the `ALL`, do we really need the `isSuperuser` flag?

---

### Known issues

-   [x] the user data store is used to store the visualization object when switching visualization from DV to Maps. Currently it relies on `d2` as it's using some classes defined in that package.
-   [x] now that with this PR DV uses the app service data for getting/setting the current AO in the user data store, Maps should be updated in order to read the AO from the same key, since the app service stores the key under `settings`. Fixed [here](https://github.com/dhis2/maps-app/pull/2247).

---

### Screenshots

Not very meaningful but this is the app running on this branch:
Screenshot 2022-07-21 at 15 16 34


edoardo:
Closing.
Superseded by #3023 .

PR #2158, opened 2022-09-27
fix: do not override height if present in style object DHIS2-13812

edoardo:
Fixes [DHIS2-13812](https://jira.dhis2.org/browse/DHIS2-13812)

---

### Key features

1. fix hidden PT plugin in dashboard

---

### Description

Dashboard app passed `height` and `width` in a `style` prop to the DV plugin.
A previous change to fix the same issue in the interpretation modal in DV app was overriding the `height` but caused the problem in dashboard app.

The fix is to use the passed value for `height` if available, and fallback to `100%`.
The former works for dashboard app while the latter works for the interpretation modal in DV.

---

### Screenshots

Before:
Screenshot 2022-09-27 at 11 52 45

After:
Screenshot 2022-09-27 at 11 51 48

### DIFF
--- a/packages/plugin/src/VisualizationPlugin.js
+++ b/packages/plugin/src/VisualizationPlugin.js
@@ -328,8 +328,11 @@ export const VisualizationPlugin = ({
                             onDrill ? onToggleContextualMenu : undefined
                         }
                         id={id}
-                        // force height otherwise the PivotTable container sets 0 as height hiding the table content
-                        style={{ ...transformedStyle, height: '100%' }}
+                        // force height when no value available otherwise the PivotTable container sets 0 as height hiding the table content
+                        style={{
+                            ...transformedStyle,
+                            height: transformedStyle.height || '100%',
+                        }}
                     />
                 ) : (
                     <ChartPlugin


jenniferarnesen:
Is it possible to add a jest test for this? Also, I noticed that the VisualisationPlugin.spec.js has some outdated things (like the `d2` property)

packages/plugin/src/VisualizationPlugin.js
```diff
@@ -328,8 +328,11 @@ export const VisualizationPlugin = ({
                             onDrill ? onToggleContextualMenu : undefined
                         }
                         id={id}
-                        // force height otherwise the PivotTable container sets 0 as height hiding the table content
-                        style={{ ...transformedStyle, height: '100%' }}
+                        // force height when no value available otherwise the PivotTable container sets 0 as height hiding the table content
+                        style={{
+                            ...transformedStyle,
+                            height: transformedStyle.height || '100%',
```
martinkrulltott, line 334:
```suggestion
                            ...(!transformedStyle.height && { height: '100%' }),
```
Here's another way to do it, either way works though 👍

martinkrulltott (approved):

PR #2166, opened 2022-10-11
refactor: add dashboard plugin support

edoardo:
Related to [DHIS2-14280](https://dhis2.atlassian.net/browse/DHIS2-14280).

### Key features

1. add plugin wrapper component
1. configure plugin entrypoint

---

### Description

The wrapper component is basically the same as the one in LL, except for the plugin component path/name.
This wrapper can be extracted somewhere (app-shell) and reused in all apps just passing the plugin component to render.

Note about `displayProperty` and `displayNameProperty`.
The `displayProperty` is `userSettings.keyAnalysisDisplayProperty` and is the one passed to the analytics request:
https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-239/analytics.html#webapi_analytics_query_parameters

`displayNameProperty` is something we derive from the above to use in other requests.

See:
https://github.com/dhis2/data-visualizer-app/blob/feat/plugin-wrapper-for-iframe/src/components/UserSettingsProvider.js#L26-L30

---

### TODO

- [x] bump `cli-app-scripts` after alpha is merged (https://github.com/dhis2/app-platform/pull/740)

---

### Known issues

Chart redraw is not triggered on window resize in dashboard (with the new iframe approach).
Fixed in https://github.com/dhis2/data-visualizer-app/pull/2166/commits/a7e08d8c83a815f33a2150b6cec213e12fab085f and https://github.com/dhis2/data-visualizer-app/pull/2166/commits/593d350dbe25e3adefa985317e28794c9b0379a6.

Screen cast demonstrating it works:

https://user-images.githubusercontent.com/150978/201341419-09ff9f55-7fdc-49e0-8565-0af4fe789ad0.mov

[DHIS2-14280]: https://dhis2.atlassian.net/browse/DHIS2-14280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reducers/settings.js
```diff
@@ -27,6 +27,9 @@ export default (state = DEFAULT_SETTINGS, action) => {
 
 export const sGetSettings = (state) => state.settings
 
+export const sGetSettingsDisplayProperty = (state) =>
```
martinkrulltott, line 30: what's the difference between the new `displayProperty` and the existing `displayNameProperty`?
Would be great if this was explained in the PR description.

martinkrulltott (approved):

src/reducers/settings.js
edoardo: The `displayProperty` is `settings.keyAnalysisDisplayProperty` and is the one passed to the analytics request:
https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-239/analytics.html#webapi_analytics_query_parameters

`displayNameProperty` is something we derive from the above to use in other requests.

See:
https://github.com/dhis2/data-visualizer-app/blob/feat/plugin-wrapper-for-iframe/src/components/UserSettingsProvider.js#L26-L30

edoardo: Before it was doing the same thing but with a double conversion and the naming was confusing as it was storing the derived value in `displayName` and then converting it back to the original one for using it in the analytics request.

PR #2214, opened 2022-11-29
fix: type selector: Wrong hover style for active selection (DHIS2-9262)

martinkrulltott:
Implements [DHIS2-9262](https://jira.dhis2.org/browse/DHIS2-9262)

---

### Key features

1. Removes the hover style from the active list item

---

### Description

When hovering the active (green) item, the item will remain unchanged. 
Previously, the "hover" style (gray) would overwrite the "active" style (green).

Note: Can't be shown in a screenshot properly, but can be seen on the Netlify build.

---

### Screenshots

image


cooper-joe (approved):
👍

PR #2224, opened 2022-12-28
fix: error message for non-numeric data types (DHIS2-10067)

martinkrulltott:
Implements [DHIS2-10067](https://jira.dhis2.org/browse/DHIS2-10067)

---

### Key features

1. Throw error when only text dx items are used in non-PT visualizations

---

### Description

When only text dx items, such as "narrative" are used, the data returned can't be displayed in a non-PT visualization.
To prevent this, there's a new error that's thrown when this situation occurs.
Note the following cases:

_Non-PT visualizations_
- x number of text items: throw error
- x number of text items + at least 1 numeric item: display visualiation

_PT visualization_
- always display visualization, regardless of item / types

---

### TODO

-   [x] Tests?

---

### Screenshots

_using only a text item throws the error_

image

_using a text item and a valid item doesn't throw the error_

image

_text items in PT can still be used_

image

[DHIS2-10067]: https://dhis2.atlassian.net/browse/DHIS2-10067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/error.js
```diff
@@ -323,6 +323,18 @@ export class NoOrgUnitResponseError extends VisualizationError {
     }
 }
 
+export class DataTypeError extends VisualizationError {
```
janhenrikoverland, line 326: "Data type" isn't really a thing, so I think we should call it `ValueTypeError`

martinkrulltott: updated

janhenrikoverland:
@martinkrulltott Looks good. Makes sense to add a test for this.

janhenrikoverland (approved):
LGTM

PR #2226, opened 2023-01-06
fix: update design to match Line Listing app

cooper-joe:
Implements [UX-130](https://dhis2.atlassian.net/browse/UX-130)

---

### Description

Updates Data Visualizer color, spacing, and typography usage to better match the updated styles defined by Line Listing app. 

---

### Screenshots

Before:

image

After:

image

[UX-130]: https://dhis2.atlassian.net/browse/UX-130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/DimensionsPanel/DndDimensionsPanel.js
+++ b/src/components/DimensionsPanel/DndDimensionsPanel.js
@@ -19,15 +19,20 @@ export class DndDimensionsPanel extends Component {
     render() {
         return (
             <div className={styles.container} data-test="dimensions-panel">
-                <DimensionFilter
-                    placeholder={i18n.t('Filter dimensions')}
-                    text={this.state.filterText}
-                    onChange={this.onFilterTextChange}
-                    onClear={this.onClearFilter}
-                    disableUnderline={true}
-                    type="search"
-                    dataTest="dimensions-panel-filter"
-                />
+                <div
+                    className={styles.filter}
+                    data-test="dimensions-panel-filter"
+                >
+                    <DimensionFilter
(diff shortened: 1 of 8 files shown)


src/components/DimensionsPanel/DndDimensionsPanel.js
```diff
@@ -19,15 +19,20 @@ export class DndDimensionsPanel extends Component {
...
-                    text={this.state.filterText}
-                    onChange={this.onFilterTextChange}
-                    onClear={this.onClearFilter}
-                    disableUnderline={true}
-                    type="search"
-                    dataTest="dimensions-panel-filter"
-                />
+                <div
+                    className={styles.filter}
+                    data-test="dimensions-panel-filter"
```
martinkrulltott, line 24:
```suggestion
```
`dimensions-panel-filter` is already passed to `DimensionFilter` below so this shouldn't be set on the `div` as well.

martinkrulltott: For reference, `dataTest` / `data-test` is only there for our Cypress tests to find elements

martinkrulltott (approved):
The code LGTM, except for a redundant `data-test` prop that should be removed.

cooper-joe:
> The code LGTM, except for a redundant data-test prop that should be removed.

Removing that `data-test` attribute seems to have caused 390 cypress tests to fail. Before that change, everything passed. Any idea what broke?

PR #2235, opened 2023-02-08
fix: use app runtime's datastore service

edoardo:
### Key features

1. replace the old `d2` based user data store with the new one

---

### Description

Maps app swapped the old `d2` implementation with the new datastore service from `app-runtime`.
This means that the `View as Map/chart` functionality is not working anymore between Maps and DV until this PR is merged.
The new service use a different path for the data store keys.


turban (approved):
LGTM! 

Nice trick to use hooks in a class component.

jenniferarnesen (approved):

PR #2223, opened 2022-12-27
feat: single value background color change based upon legend (DHIS2-13702)

martinkrulltott:
Implements [DHIS2-13702](https://jira.dhis2.org/browse/DHIS2-13702)

**Requires https://github.com/dhis2/analytics/pull/1402**

---

### Key features

1. Adds support for background color from legend for SV

---

### Description

Adds the legend style options in the options modal for SV legends to enable the new SV background color feature in Analytics.

Adds Cypress tests for the feature. As this feature affects title / subtitle colors and there were previously no tests for changing font style colors in general these tests were added as well (to `fontStyles`).

---

### Screenshots

_Legend options (same as `pivot table`, `gauge` etc)_

image

_For the actual visualization, see [Analytics](https://github.com/dhis2/analytics/pull/1402)_

[DHIS2-13702]: https://dhis2.atlassian.net/browse/DHIS2-13702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2236, opened 2023-02-17
feat: implement data icon option for SV visualization (DHIS2-10496)

edoardo:
Implements [DHIS2-10496](https://jira.dhis2.org/browse/DHIS2-10496)

Requires: https://github.com/dhis2/analytics/pull/1440

---

### Key features

1. implement option for showing data icon in SV

---

### Description

In maintenance app is possible to assign an icon to a data element, indicator, etc...
If an icon is assigned, when using the SV visualization type, it should be possible to see the icon together with the value.
This should be controlled via the options dialog.

---

### Screenshots

Option in the Style tab for SV only:
Screenshot 2023-02-17 at 14 11 21

Example of SV with icon:
Screenshot 2023-03-02 at 10 53 19

[DHIS2-10496]: https://dhis2.atlassian.net/browse/DHIS2-10496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/VisualizationOptions/Options/DataIcon.js
+++ b/src/components/VisualizationOptions/Options/DataIcon.js
@@ -1,0 +1,17 @@
+import i18n from '@dhis2/d2-i18n'
+import React from 'react'
+import { CheckboxBaseOption } from './CheckboxBaseOption.js'
+
+const DataIcon = () => (
+    <CheckboxBaseOption
+        label={i18n.t('Show data item icon')}
+        helpText={i18n.t(
+            'If the data item has an icon, display it next to the value'
+        )}
+        option={{
+            name: 'icons',
+        }}
+    />
+)
+
+export default DataIcon
--- a/src/components/VisualizationPlugin/VisualizationPlugin.js
+++ b/src/components/VisualizationPlugin/VisualizationPlugin.js
@@ -152,6 +152,34 @@ export const VisualizationPlugin = ({
                 forDashboard
             )
 
+            // parse responses to extract dx ids from metaData
+            // multiple responses are only for YOY which does not support legends or custom icon
+            // safe to use only the 1st
+            // dx dimensions might not be present, the empty array covers that case
+            const dxIds = responses[0].metaData.dimensions.dx || []
+
+            // DHIS2-10496: show icon on the side of the single value if an icon is assigned in Maintenance app and
+            // the "Show data item icon" option is set in DV options
+            if (
+                Boolean(visualization.icons?.length) &&
+                dxIds[0] &&
+                responses[0].metaData.items[dxIds[0]].style?.icon
+            ) {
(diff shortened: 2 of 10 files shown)


src/components/VisualizationOptions/Options/DataIcon.js
```diff
@@ -0,0 +1,17 @@
...
+import { CheckboxBaseOption } from './CheckboxBaseOption.js'
+
+const DataIcon = () => (
+    <CheckboxBaseOption
+        label={i18n.t('Show data item icon')}
+        helpText={i18n.t(
+            'If the data item has an icon, display it next to the value'
+        )}
+        option={{
+            name: 'icons',
```
martinkrulltott, line 12: is `icons` coming from the backend? if not, maybe we can change it to something more descriptive like `showIcon`  (or even more detailed like `showDataItemIcon` if that makes sense) in case there are more icon related options coming in the future

martinkrulltott (approved):

src/components/VisualizationOptions/Options/DataIcon.js
edoardo: Yes, it is called `icons` in the backend.
I also thought maybe to use a more explicative name, but for the other options it seems we just keep the original name in the AO, which it makes sense for readability and debugging once you know that the names in the options correspond to the ones you see in the AO.

martinkrulltott (changes requested):
Needs Cypress tests for the Analytics part

martinkrulltott:
Do not merge until https://github.com/dhis2/data-visualizer-app/pull/2223 is merged and the icons have been tested in combination with the background color as well.

src/components/VisualizationPlugin/VisualizationPlugin.js
```diff
@@ -152,6 +152,34 @@ export const VisualizationPlugin = ({
...
+            // safe to use only the 1st
+            // dx dimensions might not be present, the empty array covers that case
+            const dxIds = responses[0].metaData.dimensions.dx || []
+
+            // DHIS2-10496: show icon on the side of the single value if an icon is assigned in Maintenance app and
+            // the "Show data item icon" option is set in DV options
+            if (
+                Boolean(visualization.icons?.length) &&
+                dxIds[0] &&
+                responses[0].metaData.items[dxIds[0]].style?.icon
```
martinkrulltott, line 166:
```suggestion
                responses[0].metaData.items[dxIds[0]]?.style?.icon
```

martinkrulltott: could this ever fail? add `?` just in case?

martinkrulltott (approved):

PR #2238, opened 2023-02-28
fix: use app runtime's datastore service v39

edoardo:
Backport of https://github.com/dhis2/data-visualizer-app/pull/2235.

Fixes: https://dhis2.atlassian.net/browse/DHIS2-14841

### Key features

1. replace the old `d2` based user data store with the new one

---

### Description

Maps app swapped the old `d2` implementation with the new datastore service from `app-runtime`.
This means that the `View as Map/chart` functionality is not working anymore between Maps and DV until this PR is merged.
The new service use a different path for the data store keys.


turban (approved):
LGTM. Will do some extra testing on play 2.39 after this is merged.

PR #2240, opened 2023-03-09
fix: remove old i18n

edoardo:
### Key features

1. do not initialise the old i18n with property files

---

### Description

i18n properties files have been removed already, but the `D2Shim` was still using `i18nRoot` which causes the app to try to load the property file for the current locale.

---

### Screenshots

Failed request for old i18n property file:
Screenshot 2023-03-09 at 15 34 30


edoardo:
Closing because I created this branch from `master` instead of `dev`.

PR #2321, opened 2023-05-25
test: prevent items in transfer from being stale when filtering

martinkrulltott:
There's been a long running issue with the `Transfer` component showing stale items after applying a filter. This seems to be due to the way Cypress uses cache to store an element (which makes it stale when the element content has changed).

This solution hopefully fixes this for good, by switching to a different method for evaluating the list of items in the `Transfer` component.


HendrikThePendric (approved):
LGTM!

PR #2322, opened 2023-05-25
fix: inject style in props passed to VisualizationPlugin (DHIS2-15126)

edoardo:
Fixes [DHIS2-15126](https://dhis2.atlassian.net/browse/DHIS2-15126)

### Key features

1. pass `style` to `VisualizationPlugin`

---

### Description

`style` is used to compute the correct width when toggling the legend panel in dashboard.
`style` is not passed in props from dashboard because doing so can cause re-rendering of the whole iframe tag which causes refetching of the plugin and everything else (analytics included) (see: https://github.com/dhis2/dashboard-app/pull/2247).
Instead, `style` can be computed directly in the `PluginWrapper` based on the parent element dimensions, these should match with `width` and `height` set on the `iframe` tag.

---

### Known issues

1. ~browser window or dashboard item resize causes the legend to remain in the same position, resulting in either too much space on the right side or cutoff~ Fixed with the resize observer

---

### Screenshots

Before:
Screenshot 2023-05-25 at 15 33 34

After:
Screenshot 2023-05-25 at 15 33 52

[DHIS2-15126]: https://dhis2.atlassian.net/browse/DHIS2-15126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/VisualizationPlugin/VisualizationPlugin.js
```diff
@@ -41,6 +41,26 @@ export const VisualizationPlugin = ({
     const [contextualMenuConfig, setContextualMenuConfig] = useState({})
     const [showLegendKey, setShowLegendKey] = useState(false)
     const [renderId, setRenderId] = useState(id)
+    const [size, setSize] = useState({ width: 0, height: 0 })
+
+    const containerRef = useCallback((node) => {
```
HendrikThePendric, line 46: Perhaps since this is a [callback ref](https://legacy.reactjs.org/docs/refs-and-the-dom.html#callback-refs), and not just a regular reference to a DOM element, it could be better to reflect this in the variable name, i.e. `containerCallbackRef` of `onContainerMounted` (not so sure about that one)

HendrikThePendric (approved):
I left one remark about variable naming, which is no biggie. Other than that, this looks like a good solution to me 👍 .

PR #2358, opened 2023-05-30
feat: use Toolbar and ToolbarSidebar from analytics

HendrikThePendric:
Implements [DHIS2-15167](https://jira.dhis2.org/browse/DHIS2-15167)

**Requires https://github.com/dhis2/analytics/pull/1478**

---

### Key features

1. Updated toolbar with hover menu menu

---

### TODO

-   [x] Update `@dhis2/analytics` once released.
-   [x] Update tests for the new implementation

---

### Screenshots

dv_demo

[DHIS2-15167]: https://dhis2.atlassian.net/browse/DHIS2-15167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -10,6 +10,7 @@
         "build": "d2-app-scripts build",
         "start": "d2-app-scripts start",
         "start:nobrowser": "BROWSER=none yarn start",
+        "start:localanalytics": "cd ../analytics && yarn build && cd ../data-visualizer-app && rm -rf node_modules/@dhis2/analytics/build && cp -R ../analytics/build/ node_modules/@dhis2/analytics/build && yarn start",
```
edoardo, line 13: Sometimes this alone doesn't work due to `node_modules/.cache` removing that folder helps.

edoardo (approved):

src/components/VisualizationTypeSelector/styles/VisualizationTypeSelector.module.css
```diff
@@ -3,17 +3,29 @@
     margin-left: auto;
 }
 
+.arrowIcon.listIsOpen {
+    transform: rotate(180deg);
+}
```
HendrikThePendric, line 8: I'm now rotating the arrow-icon when the list is open, as we do on select elements to give more visual feedback to the user and make the UI more consistent.

src/components/VisualizationTypeSelector/styles/VisualizationTypeSelector.module.css
```diff
@@ -3,17 +3,29 @@
...
+
+.button:hover,
+.button.listIsOpen {
+    background-color: var(--colors-grey200);
 }
 
-.button > span {
+.button > .selectedVizTypeLabel {
     font-size: 14px;
+    padding-top: 1px;
```
HendrikThePendric, line 27: This padding is problematic. I'm not actually sure if it adds any value moving the text down here. In the LL I did a similar fix, which does make sense because we always show the same icon. However, in DV each viz-type comes with its own icon. In some cases things now look better aligned, but for other icons this is not the case.

PR #2368, opened 2023-06-01
test: extend timeout for loading visualizations

martinkrulltott:
extends the timeout for when visualizations are loading since that seems to be slow sometimes (which could be because of the server), e.g. this test failed when trying to open a PR because the result didn't return fast enough https://cloud.cypress.io/projects/sojh88/runs/2000/test-results/834f2e45-c119-4b4f-9c08-d7f95653be2b


janhenrikoverland (approved):

HendrikThePendric (approved):
40s is quite a long wait, but if that's what it takes 👍

PR #2613, opened 2023-07-06
chore: upgrade cypress to v12

HendrikThePendric:
Implements an upgrade to Cypress 12

---

### Key features

1. Allows us to use Cypress 12

---

### Description

I have basically implemented the same upgrade as I did here https://github.com/dhis2/line-listing-app/pull/374

---

### Gotcha

If you get stuck on the login screen, or if the app never loads in `cy:open` this could be caused be the service-worker. To fix that, just open DevTools in the Cypress window and go the >>Application>>Storage>>Clear site data


HendrikThePendric:
NOTE: I've taken this PR out of draft and added the "On hold" label, because if the PR is in draft, the e2e checks are skipped.

HendrikThePendric:
The 3 test failures we're seeing here all seem to be related to server issues to me, and @martinkrulltott confirmed this. So I think we could say the upgrade has been successful and the changes in the current PR are correct.
But we still cannot merge 🤷‍♂️

martinkrulltott (approved):

PR #2746, opened 2023-08-25
feat: cumulativeValues in PT DHIS2-5497

edoardo:
Implements [DHIS2-5497](https://jira.dhis2.org/browse/DHIS2-5497)

**Requires https://github.com/dhis2/analytics/pull/1567**

---

### Key features

1. add `cumulativeValues` option in PT
2. disable `Totals`, `Number type` and `Use legend for chart colors` options when `cumulativeValues` is checked
3. add helper texts on options to explain the behaviours
4. added Cypress helper functions and test for the cumulativeValues option

---

### Description

The `cumulativeValues` option is already implemented for other vis types.
We reuse the same option for PT so we don't need any change in the backend, but it requires some custom behaviours because in PT when `cumulativeValues` is enabled other options don't make sense and should be disabled.

---

### Known issues

-   [x] when totals and `numberType` options are disabled, they are still applied to the PT table if their values are initialised (checkboxes checked and select with a selection).

---

### Screenshots

Data tab in Options modal with `cumulativeValues` unchecked for PT vis type:
Screenshot 2023-08-25 at 16 05 06

Data tab in Options modal with `cumulativeValues` checked for PT vis type:
Screenshot 2023-11-13 at 14 31 39

Legend tab in Options modal with `cumulativeValues` checked for PT vis type:
Screenshot 2023-11-13 at 14 31 45
Screenshot 2023-11-13 at 14 32 02

[DHIS2-5497]: https://dhis2.atlassian.net/browse/DHIS2-5497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reducers/ui.js
```diff
@@ -728,6 +737,12 @@ export const sGetAxisSetup = (state) => {
         : []
 }
 
+export const sGetUiDisabledOption = (state, option) => {
+    const disabledOptions = sGetUiDisabledOptions(state)
+
+    return disabledOptions[option.name]
+}
+
```
janhenrikoverland, line 745: Maybe just do `=> sGetUiDisabledOptions(state)[option.name]`

src/modules/disabledOptions.js
```diff
@@ -0,0 +1,37 @@
...
+                    ;[
+                        'colTotals',
+                        'colSubTotals',
+                        'rowTotals',
+                        'rowSubTotals',
+                        'numberType',
+                    ].forEach((option) => delete disabledOptions[option])
+                }
+
+                break
```
janhenrikoverland, line 31: If another `case 'someOtherOption'` is added later this break could become a sneaky bug :smile: . Maybe add a comment or remove it right away. Performance gain is very small. If I understand it correctly though.

src/reducers/ui.js
edoardo: Changed.

janhenrikoverland (approved):

PR #2872, opened 2023-10-13
test: break a test with adel

martinkrulltott:
just testing how the Cypress dashboard reports failing tests
remove this PR and branch after today


martinkrulltott:
this was only for testing purposes, so will delete this now

PR #3000, opened 2024-03-20
fix: add classnames and update json instructions file for outlier tables

HendrikThePendric:
Implements [DHIS2-17007](https://dhis2.atlassian.net/browse/DHIS2-17007)

**Requires https://github.com/dhis2/push-analytics/pull/13**

---

### Key features

1. Adds support for outlier tables in push analytics

---

### Description

1. Most of the changes in this PR are directly related to the outlier tables
2. But I have also added a class-name to the start screen so it is clearer for push-analytics to know when the visualization was cleared

[DHIS2-17007]: https://dhis2.atlassian.net/browse/DHIS2-17007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/public/push-analytics.json
+++ b/public/push-analytics.json
@@ -39,10 +39,14 @@
                             "YEAR_OVER_YEAR_LINE",
                             "YEAR_OVER_YEAR_COLUMN",
                             "SCATTER",
-                            "BUBBLE",
-                            "SINGLE_VALUE"
+                            "BUBBLE"
                         ],
                         "selector": ".highcharts-container > svg"
+                    },
+                    {
+                        "dashboardItemProperty": "visualization.type",
+                        "value": "SINGLE_VALUE",
+                        "selector": "[data-test='visualization-container']"
                     }
                 ]
             }
--- a/public/push-analytics.json
+++ b/public/push-analytics.json
@@ -93,7 +93,7 @@
             "dashboardItemProperty": "visualization.type",
             "value": "OUTLIER_TABLE",
             "strategy": "interceptRequest",
-            "urlReg": "*/analytics/outlierDetection?displayProperty=NAME&*"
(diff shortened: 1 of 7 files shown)


public/push-analytics.json
```diff
@@ -53,6 +89,12 @@
             "htmlSelector": "body",
             "cssSelector": "style"
         },
+        {
+            "dashboardItemProperty": "visualization.type",
+            "value": "OUTLIER_TABLE",
+            "strategy": "interceptResponse",
+            "urlGlob": "**/analytics/outlierDetection?displayProperty=NAME&**"
```
edoardo, line 96: This argument might have a different value depending on some settings.
In that case this won't match, perhaps remove the value or the entire query string?

public/push-analytics.json
```diff
@@ -28,23 +40,47 @@
                             "YEAR_OVER_YEAR_COLUMN",
                             "SCATTER",
                             "BUBBLE",
-                            "SINGLE_VALUE",
-                            "PIVOT_TABLE",
-                            "OUTLIER_TABLE"
+                            "SINGLE_VALUE"
                         ],
-                        "selector": ".highcharts-container"
+                        "selector": ".highcharts-container > svg"
```
edoardo, line 45: I'm not sure I understand this correctly, but I don't think this selector is present for SINGLE_VALUE?
Single value does not use Highcharts for rendering the SVG.

edoardo (approved):

public/push-analytics.json
HendrikThePendric: Good point, solved now!

PR #2892, opened 2023-10-26
fix: avoid undefined in totals (DHIS2-14511) [v39]

edoardo:
Fixes [DHIS2-14511](https://jira.dhis2.org/browse/DHIS2-14511)

**Requires https://github.com/dhis2/analytics/pull/1585**

---

### Key features

1. fix undefined in PT totals when value is 0 (bump analytics dependency)

---

Test editing the description
- [x] Test editing the dscription with checkbox

### Description

Backport.
See original fix [here](https://github.com/dhis2/analytics/pull/1552).

[DHIS2-14511]: https://dhis2.atlassian.net/browse/DHIS2-14511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3003, opened 2024-03-20
fix: about AO showed wrong datetime, newlines in interpretations and replies not preserved [v39]

jenniferarnesen:
Fixes https://dhis2.atlassian.net/browse/DHIS2-15825 and https://dhis2.atlassian.net/browse/DHIS2-16365

Requires https://github.com/dhis2/analytics/pull/1642

This PR supersedes https://github.com/dhis2/data-visualizer-app/pull/2892

Key features
1. Show correct date and time in About AO Unit in interpretations panel
2. Preserve newlines in interpretations and replies
(yarn-deduplicate was run to make sure that plugin also uses the 24.10.3 version of analytics.)

Other changes since analytics was last updated:
1. feat: single value background color change based upon legend (DHIS2-13702)
2. fix: sv size when downloading is a small square (DHIS2-15178)
3. fix: avoid undefined in totals DHIS2-14511

Screenshot shows chart was just created. Timestamp is "a few seconds ago", rather than "1 hour ago" which is the bug you would see in GMT+1
image
image

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -13,7 +13,7 @@
         "redux-mock-store": "^1.5.4"
     },
     "dependencies": {
-        "@dhis2/analytics": "^24.9.3",
+        "@dhis2/analytics": "^24.10.3",
         "@dhis2/app-runtime": "^3.9.0",
         "@dhis2/app-runtime-adapter-d2": "^1.1.0",
         "@dhis2/app-service-datastore": "^1.0.0-beta.3",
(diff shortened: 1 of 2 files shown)


edoardo (approved):

martinkrulltott:
It's been a long time since I looked at it last, but the two tickets from me (DHIS2-13702, DHIS2-15178) seems to be working as they should after some brief testing.

PR #3017, opened 2024-04-11
fix: plugin flashes when interacting with Interpretations modal [DHIS2-15570]

jenniferarnesen:
Fixes [DHIS2-15570](https://dhis2.atlassian.net/browse/DHIS2-15570)
Fixes [DHIS2-17137](https://dhis2.atlassian.net/browse/DHIS2-17137)

**Requires https://github.com/dhis2/analytics/pull/1608**

---

### Key features

1. Prevent re-requests for analytics, and resulting re-renders that were giving a flashing effect.

---

### Description

1. InterpretationModal: memoize the callback to avoid a re-render due to a "new" function being passed with each render of InterpretationModal. The effect of this change is minor, probably just 1 rerender
2. VisualizationPlugin: ensure that the size of the plugin is set when it is coming from the modal so that when changing the width of the viewport, the visualization adjusts its size accordingly. (Previously this "worked" but only because of all the re-fetch/re-renders)
3. VisualizationPluginWrapper: need to listen to the individual properties, rather than the whole props object since that was triggering lots of unnecessary re-fetch/renders.
4. set the max-height to respect viewport so that vertical resize causes the chart to adjust accordingly
 
### Video
before (width):

https://github.com/dhis2/data-visualizer-app/assets/6113918/7edc8e3f-88a9-4794-a122-4a7012e2dcc3

before (height):

https://github.com/dhis2/data-visualizer-app/assets/6113918/dd3cc08f-0f54-4ba9-a8a2-4311e39fd83a

after (width):
https://github.com/dhis2/data-visualizer-app/assets/6113918/5dbeb7e4-edcd-4f5f-9b4a-e829c9061920

after (height):

https://github.com/dhis2/data-visualizer-app/assets/6113918/ac640b3c-4149-4159-8257-a6915ba91b5c

[DHIS2-15570]: https://dhis2.atlassian.net/browse/DHIS2-15570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-17137]: https://dhis2.atlassian.net/browse/DHIS2-17137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/VisualizationPlugin/VisualizationPluginWrapper.js
```diff
@@ -27,10 +28,39 @@ const VisualizationPluginWrapper = (props) => {
...
+    useEffect(() => {
+        setPluginProps({
+            displayProperty: props.displayProperty,
+            visualization: props.visualization,
+            filters: props.filters,
+            forDashboard: props.forDashboard,
+            id: props.id,
+            isInModal: props.isInModal,
+            style: props.style,
+            onChartGenerated: props.onChartGenerated,
```
edoardo, line 40: I don't think all these callbacks are needed or used by dashboard/interpretations modal.
Perhaps we should keep this to the minimum required to avoid confusion?

PR #3023, opened 2024-04-18
chore: remove d2

edoardo:
Implements [DHIS2-19197](https://jira.dhis2.org/browse/DHIS2-19197)

---

### Key features

1. get rid of `d2` entirely

---

### Description

This branch has changes to remove the need for `d2`.

---

### TODO

- [x] figure out how to use the user data store without `d2`
- [x] lots of regression testing
- [x] fix tests, particularly the App.js tests, but before touching them we should figure out how we want to test, what should 
be tested in Cypress and what in unit tests
- [x] look for and clean up more code not needed
- [x] see if it's possible to avoid fetching all user authorities to check for the `ALL`, do we really need the `isSuperuser` flag?
- [x] fix DimensionsPanel.spec.js: figure out how to mock the cachedDataQuery / provider (tests are skipped)
- [x] tester approved (Jennifer)

---

### Known issues

-   [x] the user data store is used to store the visualization object when switching visualization from DV to Maps. Currently it relies on `d2` as it's using some classes defined in that package.
-   [x] now that with this PR DV uses the app service data for getting/setting the current AO in the user data store, Maps should be updated in order to read the AO from the same key, since the app service stores the key under `settings`. Fixed [here](https://github.com/dhis2/maps-app/pull/2247).

---

### Screenshots

Not very meaningful but this is the app running on this branch:
Screenshot 2022-07-21 at 15 16 34

[DHIS2-19197]: https://dhis2.atlassian.net/browse/DHIS2-19197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HendrikThePendric (approved):
These changes all look great, the bundle size will have decreased and I think the code is a lot easier to follow in many places now. The changes we discussed when you walked me through the changed have all been implemented, so I think this is ready for KFMT now 🚀 .

HendrikThePendric:
Just one additional note after my approval. If fixing the SonarCube issue is easy to do then it would be good to address it. 

But on the other hand:
- The level of complexity was already present, you actually simplified this code.
- It's a bit of a false positive anyway, the code is indeed 4 levels but that's just because there are two nested `describe` scopes and one `it`, so the "brain overload" tag here is not really applicable.

cypress/integration/options/icon.cy.js
```diff
@@ -126,7 +127,7 @@ describe('Icon', () => {
         })
     })
     // TODO: Skipped because of the same reason as the commented out tests above
-    it.skip('icon gets correct color when a legend is in use', () => {
+    it('icon gets correct color when a legend is in use', () => {
```
jenniferarnesen, line 130: The TODO can be removed, right?

edoardo:
> Just one additional note after my approval. If fixing the SonarCube issue is easy to do then it would be good to address it.
> 
> But on the other hand:
> 
> * The level of complexity was already present, you actually simplified this code.
> * It's a bit of a false positive anyway, the code is indeed 4 levels but that's just because there are two nested `describe` scopes and one `it`, so the "brain overload" tag here is not really applicable.

I refactored the test and extracted the intercept function.

cypress/integration/dimensions/period.cy.js
```diff
@@ -487,4 +465,77 @@ describe('Period dimension', () => {
...
+            cy.intercept(
+                /systemSettings/,
+                systemSettingsInterceptFn({
+                    ...systemSettingsBody,
+                    keyHideBiWeeklyPeriods: true,
+                })
+            )
+            goToStartPage()
+        })
+        it('opens the period dimension modal', () => {
```
jenniferarnesen, line 479: Cypress testing best practices would suggest that all this should be in a single test:
https://docs.cypress.io/app/core-concepts/best-practices#Having-Tests-Rely-On-The-State-Of-Previous-Tests
https://docs.cypress.io/app/core-concepts/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion

edoardo: Refactored.

HendrikThePendric (approved):

jenniferarnesen (approved):

PR #3031, opened 2024-04-23
fix: accept displayProperty of both with and without display prefix [v39]

jenniferarnesen:
Part of fix for https://dhis2.atlassian.net/browse/DHIS2-12499

The LL and DV plugin treat the `displayProperty` prop differently. LL accepts the value as is, while DV expects it to have the 'display' prefix and then converts it. The result is that displayProperty does not get set for DV for the analytics request. 

Before:
image

After:


jenniferarnesen:
not needed

PR #3033, opened 2024-04-25
fix: single value subtext was not displayed correctly [v39]

jenniferarnesen:
Fixes https://dhis2.atlassian.net/browse/DHIS2-13077


jenniferarnesen:
superseded

PR #3073, opened 2024-05-31
feat: improve support for right-to-left languages in Data Visualiser

kabaros:
Implements [DHIS2-16778](https://dhis2.atlassian.net/browse/DHIS2-16778)

### Key features

1. Use _logical_ CSS properties instead of physical properties throughout the project (done automatically with a tool)
2. Update the `cli-style` to a version that supports checking logical properties (and automatically converting physical to logical properties)

---

### Description

This PR implements better support for right-to-left languages in the app, by leveraging the use of [logical CSS properties](https://css-tricks.com/css-logical-properties-and-values/ ) that are independent of language direction, rather than physical properties (i.e. `*-start` instead of `-left` and `*-end` instead `*-right`). Logical properties are widely [supported](https://caniuse.com/css-logical-props) now and they should become the default way for layout going-forward to ensure that we support RTL languages without extra development burden (which will be enforced by `d2 style`).

Although the change set is big. It's mostly done automatically using `d2 style apply css` which under the hood uses  [stylelint](https://stylelint.io/) to enforce and apply the rules.

The better support for RTL builds on top of previous work done in the platform, namely:

- changing the UI library to support RTL: https://github.com/dhis2/ui/pull/1448
- introducing an opt-in option in `d2.config` that allows apps to declare they're ready to support RT. The platform changes also takes care of setting the correct direction and language for the header among other things: https://github.com/dhis2/app-platform/pull/825
- updates to cli-style to enforce and apply rules for using logical CSS properties: https://github.com/dhis2/cli-style/pull/464

#### Testing

- The main test here should be that the app in English (or any LTR language) should look exactly as before, after these changes.
- Test on an environment set to Arabic (or any other RTL language) to confirm it looks as expected for an RTL user.
   - This is an instance with Arabic metadata https://dev.im.dhis2.org/arabic-test-instance

---

### Screenshots

| Before | After |
| --- | --- |
| image  | image |
 
[DHIS2-16778]: https://dhis2.atlassian.net/browse/DHIS2-16778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jenniferarnesen (approved):

jenniferarnesen:
The correct netlify link is:  https://pr-3073--dhis2-data-visualizer.netlify.app/

jenniferarnesen:
> 🚀 Deployed on https://pr-3073.data-visualizer.netlify.dhis2.org

This url doesn't work. Seems like netlify has changed their url format. This one works:

https://pr-3073--dhis2-data-visualizer.netlify.app/

jenniferarnesen (changes requested):
Found some bugs

jenniferarnesen:
Bug: padding-block-end not having the intended effect in Options->Data
Bug
image
Expected
image

jenniferarnesen:
Bug: wrong font
image

jenniferarnesen:
Above reported bugs are now fixed

jenniferarnesen (approved):

PR #3131, opened 2024-07-01
fix: show error UI when plugin has empty rows [DHIS2-16793]

HendrikThePendric:
Implements [DHIS2-16793](https://jira.dhis2.org/browse/DHIS2-16793)

---

### Key features

1. Adds some validation to the plugin wrapper so that it won't attempt to render a visualization when there is no data

---

### Description

The key feature needs no further explanation but there are some things to take into consideration about the implementation:

- The JIRA issue mentions a bug in the `PivotTable` component, and this is indeed where the error is thrown. However, this is really going wrong at an earlier stage: the filter condition has caused the rows to be empty, and really the plugin shouldn't have rendered the `PivotTable` in the first place, but some sort of info/error screen instead. So this is what I have addressed, and I have ignored the bug in the `PivotTable`. I am not sure we need to address this one.
- The most common way to render these types of error views is using an `ErrorBoundary` component. I tried to implement things using one, but struggled to get this to work: somehow the `onComponentDidCatch` hook never got triggered. When I realised that a solution without an `ErrorBoundary` was very simple (less code and no complexity), more in line with the implementation in `StartScreen`, and worked at the first attempt, I decided to stick with that.
- **THIS PR CHANGES THE "NO-DATA" ON A DASHBOARD BEHAVIOUR OF ALL VISUALIZATION TYPES  AND WE NEED TO CHECK IF THIS WHAT WE WANT**

---

### TODO

-   [x] Manual testing (I don't think an e2e test is possible)
-   [ ] Verify if manual test method is correct and sufficiently covers all use cases
-   [ ] Verify if all changes are as intended

---

### Steps I took to test and illustrate the generic changes

1. I created a dashboard with the following content:
    1. **Chart**: ANC: 1-3 trend lines last 12 months
    2. **Pivot Table**: ANC: ANC 1st and 2nd visits at facilities with hierarchy
    3. **Single Value**: BCG coverage last 12 months - Bo
2. I then applied a filter filter for a future period (january 2025) to that dashboard to force missing data. I then took a screenshot.
3. I visited the data visualizer app and opened each of the visualisations above. For each one I changed the period dimension to the same future period I used on the dashboard filter and I then took a screenshot.
4. I then installed a custom DV version, built from the current PR branch, and repeated step 2 & 3.

Below I will illustrate the outcome:

Dashboard view with missing data with BEFORE the changes:
current_dashboard

Dashboard view with missing data with AFTER the changes:
after_dashboard

Data Visualizer views with no data (the was no change between BEFORE and AFTER):
**Chart**: ANC: 1-3 trend lines last 12 months
current_chart

**Pivot Table**: ANC: ANC 1st and 2nd visits at facilities with hierarchy
current_pivot_table

**Single Value**: BCG coverage last 12 months - Bo
current_single_value

So my current thinking is that we want these generic changes, because we want the consistency between the behaviour of app VS plugin. But I could be wrong. Specifically I am not 100% sure if "applying a filter on a dashboard" and "changing a dimension in DV" are the same type of thing / should result in the same UI....

### Bugfix reproduction

I also took some steps to reproduce the bug described in [DHIS2-16793](https://jira.dhis2.org/browse/DHIS2-16793) and this resulted in the following screenshot which confirms the bug is indeed gone.

Screenshot 2024-07-01 at 10 47 18

[DHIS2-16793]: https://dhis2.atlassian.net/browse/DHIS2-16793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Visualization/Visualization.js
+++ b/src/components/Visualization/Visualization.js
@@ -1,8 +1,4 @@
-import {
-    DIMENSION_ID_DATA,
-    VIS_TYPE_OUTLIER_TABLE,
-    VIS_TYPE_PIVOT_TABLE,
-} from '@dhis2/analytics'
+import { DIMENSION_ID_DATA, VIS_TYPE_PIVOT_TABLE } from '@dhis2/analytics'
 import debounce from 'lodash-es/debounce'
 import PropTypes from 'prop-types'
 import React, { Component, Fragment } from 'react'
@@ -16,10 +12,10 @@ import {
     acSetUiDataSorting,
     acAddParentGraphMap,
 } from '../../actions/ui.js'
+import { ensureAnalyticsResponsesContainData } from '../../modules/analytics.js'
 import {
     AssignedCategoriesDataElementsError,
     GenericServerError,
-    EmptyResponseError,
     AssignedCategoriesAsFilterError,
     MultipleIndicatorAsFilterError,
     NoDataOrDataElementGroupSetError,
@@ -29,7 +25,6 @@ import {
     ValueTypeError,
     AnalyticsGenerationError,
     AnalyticsRequestError,
-    NoOutliersError,
 } from '../../modules/error.js'
 import { removeLastPathSegment } from '../../modules/orgUnit.js'
 import { sGetCurrent } from '../../reducers/current.js'
@@ -144,15 +139,10 @@ export class UnconnectedVisualization extends Component {
 
         this.props.addMetadata(forMetadata)
 
-        if (
-            !responses.some((response) => response.rows && response.rows.length)
-        ) {
-            if (this.props.visualization.type === VIS_TYPE_OUTLIER_TABLE) {
-                throw new NoOutliersError()
-            }
-
-            throw new EmptyResponseError()
-        }
+        ensureAnalyticsResponsesContainData(
+            responses,
+            this.props.visualization.type
+        )
     }
 
     onDrill = (drillData) => {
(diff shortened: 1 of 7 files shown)


edoardo (approved):

src/components/Visualization/Visualization.js
```diff
@@ -144,15 +139,10 @@ export class UnconnectedVisualization extends Component {
...
-        if (
-            !responses.some((response) => response.rows && response.rows.length)
-        ) {
-            if (this.props.visualization.type === VIS_TYPE_OUTLIER_TABLE) {
-                throw new NoOutliersError()
-            }
-
-            throw new EmptyResponseError()
-        }
+        ensureAnalyticsResponsesContainData(
```
jenniferarnesen, line 142: I see that this wasn't previously in a try/catch either. Does this ever throw? What does the user experience in this case?

HendrikThePendric: I worked on this quite a while ago, but I am pretty sure it is like this (quite complicated):
1. `onResponsesReceived` is defined in `Visualization` and calls `ensureAnalyticsResponsesContainData`. It is passed as a prop to `PluginWrapper`.
2. It is and then called by the `PluginWrapper` when a response comes in: it is called by `doFetchData` which in turn is called in `doFetchAll`.
3. `doFetchAll()` returns a Promise and the catch method of this promise is called (`doFetchAll().catch(..)`). In the catch calback `onError()` is being called with the error. So if `ensureAnalyticsResponsesContainData()` throws an error this will be the parameter passed to `onError(error)`
4. `onError` itself is another prop called by the `PluginWrapper` but defined in the `Visualization`. It ends up calling `this.props.setLoadError(error)` and this will ensure the `StartScreen` is rendered with the appropriate message.

PR #3135, opened 2024-07-09
feat: add HighchartsChartProvider

HendrikThePendric:

**Requires https://github.com/dhis2/analytics/pull/XXX**

---

### Key features

1. _feature_

---

### Description

_text_

---

### TODO

-   [ ] Cypress tests
-   [ ] Update docs
-   [ ] Manual testing
-   [ ] _task_

---

### Known issues

-   [ ] _issue_

---

### Screenshots

_supporting text_

### DIFF
--- a/src/components/HighchartsChartProvider.js
+++ b/src/components/HighchartsChartProvider.js
@@ -1,0 +1,34 @@
+import PropTypes from 'prop-types'
+import React, { createContext, useCallback, useContext, useRef } from 'react'
+
+const throwIfUnitialized = () => {
+    throw new Error('HighchartsChartContext not yet initialized')
+}
+
+export const HighchartsChartContext = createContext({
+    getChartInstance: throwIfUnitialized,
+    setChartInstance: throwIfUnitialized,
+})
+
+export const useHighchartsChartContext = () =>
+    useContext(HighchartsChartContext)
+
+export const HighchartsChartProvider = ({ children }) => {
+    const highChartsChartRef = useRef(null)
+    const getChartInstance = useCallback(() => highChartsChartRef.current, [])
+    const setChartInstance = useCallback((chartInstance = null) => {
+        highChartsChartRef.current = chartInstance
+    }, [])
+
+    return (
+        <HighchartsChartContext.Provider
+            value={{ getChartInstance, setChartInstance }}
+        >
+            {children}
+        </HighchartsChartContext.Provider>
+    )
+}
+
+HighchartsChartProvider.propTypes = {
+    children: PropTypes.node,
+}
--- a/src/components/Visualization/Visualization.js
+++ b/src/components/Visualization/Visualization.js
@@ -37,6 +37,7 @@ import { sGetLoadError, sGetIsPluginLoading } from '../../reducers/loader.js'
 import { sGetSettingsDisplayProperty } from '../../reducers/settings.js'
 import { sGetUiRightSidebarOpen } from '../../reducers/ui.js'
 import LoadingMask from '../../widgets/LoadingMask.js'
+import { HighchartsChartContext } from '../HighchartsChartProvider.js'
 import { VisualizationPlugin } from '../VisualizationPlugin/VisualizationPlugin.js'
 import StartScreen from './StartScreen.js'
 import styles from './styles/Visualization.style.js'
@@ -104,7 +105,10 @@ export class UnconnectedVisualization extends Component {
         this.props.setLoadError(error)
     }
 
-    onChartGenerated = (svg) => this.props.setChart(svg)
+    onChartGenerated = (svg, highchartsChart) => {
+        this.props.setChart(svg)
+        this.context.setChartInstance(highchartsChart)
+    }
 
     onDataSorted = (sorting) => {
         this.props.onLoadingStart()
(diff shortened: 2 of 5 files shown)


src/components/HighchartsChartProvider.js
```diff
@@ -0,0 +1,34 @@
+import PropTypes from 'prop-types'
+import React, { createContext, useCallback, useContext, useRef } from 'react'
+
+const throwIfUnitialized = () => {
```
edoardo, line 4:
```suggestion
const throwIfUninitialized = () => {
```

or `throwIfNotInitialized` if it's more readable

src/components/Visualization/Visualization.js
```diff
@@ -104,7 +105,10 @@ export class UnconnectedVisualization extends Component {
         this.props.setLoadError(error)
     }
 
-    onChartGenerated = (svg) => this.props.setChart(svg)
+    onChartGenerated = (svg, highchartsChart) => {
+        this.props.setChart(svg)
```
edoardo, line 109: I guess this is still needed for SV, correct?

edoardo: I wonder if we can just pass `highchartsChart` and call `getSVGForExport` for `setChart` here.

src/components/VisualizationPlugin/ChartPlugin.js
```diff
@@ -41,7 +41,8 @@ const ChartPlugin = ({
                     visualizationConfig.visualization.getSVGForExport({
                         sourceHeight: 768,
                         sourceWidth: 1024,
-                    })
+                    }),
+                    visualizationConfig.visualization
```
edoardo, line 45: Perhaps we can pass only the latter to `onChartGenerated`, see comment above.

edoardo (approved):

HendrikThePendric:
I have had a meeting with team-platform and we discussed the problem from two angles:

### The general inability of specifying a script in a DB locale string

While considering the problem from this angle it became apparent that how we deal with locales in general needs some structural decisions. Is having multiple scripts per locale a use-case we want to support on a single instance, since it introduces a level of complexity? Do we need to keep distinguishing between UI and DB locale and that also introduces a level ambiguity? Etc, etc.... 

So this turned out this is a very broad topic which needs to be discussed and investigated thoroughly, and as such it will be a long time until this is resolved.

At the same the fact that our DB locales do not support script sections is also a practical problem in the field that the Uzbek instance is working around using bit of a hack.... They now use `uz` and `uz_UZ` and uses one for Latin and the other for Cyrillic DB translations. This is obviously not right, and this type of misuse of locale strings is problematic for the locale-based solution in this PR. The fact that Uzbek is only `Latn` (supported by standard PDF fonts) and `Cyrl` (supported by the base Noto font) is a happy accident which causes the current solution to actually function correctly for Uzbek.

### The inability to generate PDFs correctly for non-ASCII scripts

Viet brought it to my attention that the sever-side PDF generation endpoint is not just being used by the data-visualizer-app, but also by the reports-app. I already knew that the client-side solution in this PR only partially solved the problem because the single-value visualisations would still rely on the server side PDF generation endpoint, but now I realise that the problem is more widespread.

### Consequences

The points above makes me think that a fix to the server side endpoint would be more suitable than the client-side solution I have implemented in this PR:
- A server side fix would be generic, no app-specific fixes need, no exceptions in what does or doesn't support non-ASCII characters when exported to PDF.
- A [combined Noto font](https://github.com/satbyy/go-noto-universal/releases) exists with support for all modern scripts. We cannot use it for a client-side solution because it weighs 14MB. If we can use this font bundle on the server, then we could just use that for all locale strings.

I have some outstanding questions with Viet about the server side solution, but if what I mention about the combined Noto font is possible, then I would be in favour of closing this PR completely in favour of a server side fix.

src/modules/notoFontForLocale/index.js
```diff
@@ -0,0 +1,48 @@
...
+     * for which a 2 character ISO639-1 code also exists are
+     * normalized to the 2 character ISO639-1 code */
+    const jsLocale = new Intl.Locale(rawLanguage, { script, region })
+    const language = jsLocale.language
+
+    return (
+        findCustomNotoFont(({ scripts }) => scripts.has(script)) ??
+        findCustomNotoFont(({ languages }) => languages.has(language)) ??
+        NOTO_SANS_BASE_FONT_NAME
+    )
```
HendrikThePendric, line 25: Note that the script of the locale takes precedence over the language.

src/components/Visualization/Visualization.js
```diff
@@ -176,6 +177,20 @@ UnconnectedVisualization.propTypes = {
     onLoadingStart: PropTypes.func,
 }
 
+UnconnectedVisualization.contextType = ChartContext
+
+/* Setting these contextTypes is required for Jest/Enzyme
```
jenniferarnesen, line 182: It would be interesting to see if converting the tests to testing-library would resolve this problem, or if it is jest that is the problem?

HendrikThePendric: I've refactored this code to be more in line with the approach @edoardo took when removing d2 and now I don't need set these contextTypes anymore.

PR #3159, opened 2024-08-01
feat: pivot table as default vis type in DV (DHIS2-15833)

janhenrikoverland:
Implements [DHIS2-15833](https://jira.dhis2.org/browse/DHIS2-15833)

**Requires https://github.com/dhis2/analytics/pull/1694**

### TODO

-   [x] Cypress tests
-   Update docs - not needed
-   [x] Manual testing

---

### Screenshots

See https://github.com/dhis2/analytics/pull/1694 for video.

[DHIS2-15833]: https://dhis2.atlassian.net/browse/DHIS2-15833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/integration/options/legend.cy.js
+++ b/cypress/integration/options/legend.cy.js
@@ -6,11 +6,11 @@ import {
     VIS_TYPE_PIVOT_TABLE,
     VIS_TYPE_STACKED_COLUMN,
     VIS_TYPE_LINE,
-    visTypeDisplayNames,
     DIMENSION_ID_PERIOD,
     AXIS_ID_COLUMNS,
     VIS_TYPE_AREA,
     AXIS_ID_ROWS,
+    getDisplayNameByVisType,
 } from '@dhis2/analytics'
 import {
     expectChartTitleToBeVisible,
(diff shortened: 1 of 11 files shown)


cypress/integration/options/legend.cy.js
```diff
@@ -464,7 +465,7 @@ describe('Options - Legend', () => {
 
         cy.log('navigates to the start page and adds data items')
         goToStartPage()
-        changeVisType(visTypeDisplayNames[VIS_TYPE_PIVOT_TABLE])
+        changeVisType(getDisplayNameByVisType(VIS_TYPE_PIVOT_TABLE))
```
edoardo, line 468: I guess this is not needed as it should be already Pivot table by default, same for the other ones below.
Not a problem leaving this step anyway.

edoardo (approved):

cypress/integration/options/legend.cy.js
janhenrikoverland: Yep, then we avoid having the same problem again later.

PR #3218, opened 2024-09-23
chore: fix local jest run

HendrikThePendric:
This PR is not related to a JIRA issue, but the fact that it's currently impossible to run tests locally is highly problematic when trying to debug test failures after making changes in the codebase.

---

### Key features

1. Get the tests to run locally by disabling the report-portal config locally
2. Prevent console.errors from showing up in the test output (FYI these errors are not making the tests fail, but are making the test output hard to interpret)

---

### Description

The current state of the PR is captured in the TODO list

---

### TODO

-   [X] Disable report portal config locally
-   [x] Prevent component import errors (`Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.`). ~Not 100% sure how to fix this.~ **I did end up finding a good fix for this in the end**
-   [x] Prevent axis is not iterable error (`Error: Uncaught [TypeError: axis is not iterable (cannot read property undefined)]`. This is caused by `convertOuLevelsToUids` which is imported from `@dhis2/analytics` and is ultimately coming from `layoutGetAllDimensions`, which doesn't handle `axis` being undefined. Not sure if the correct fix is to adjust the test data or the implementation in `@dhis2/analytics`


src/components/VisualizationPlugin/__tests__/VisualizationPlugin.spec.js
```diff
@@ -8,6 +8,19 @@ import * as options from '../../../modules/options.js'
...
+    const noop = () => {}
+    return {
+        suppress() {
+            console.error = noop
+        },
+        restore() {
+            console.error = originalConsoleError
+        },
+    }
+})()
```
HendrikThePendric, line 22: This helper allows us to temporarily suppress console errors. It works, but whether or not we should use it is another matter....

src/components/VisualizationPlugin/__tests__/VisualizationPlugin.spec.js
```diff
@@ -167,15 +184,19 @@ describe('VisualizationPlugin', () => {
...
+            consoleErrorSuppressor.suppress()
             // eslint-disable-next-line no-import-assign, import/namespace
             api.apiFetchAnalytics = jest.fn().mockRejectedValue('error')
 
             await canvas()
 
+            consoleErrorSuppressor.restore()
+
             expect(defaultProps.onError).toHaveBeenCalled()
         })
```
HendrikThePendric, line 196: I guess that here you could argue that it's OK to suppress the console error. There are more elegant ways to assert errors in Jest, like [this one](https://jestjs.io/docs/expect#rejects), but I wouldn't know how you could then also assert that `defaultProps.onError` has been called.....

config/testSetup.js
```diff
@@ -3,3 +3,22 @@ import Enzyme from 'enzyme'
...
+global.console.error = function () {
+    const isMessageToIgnore =
+        arguments[0] === INVALID_ARG_0 &&
+        arguments[1] === INVALID_ARG_1 &&
+        arguments[2] === INVALID_ARG_2
+
+    if (!isMessageToIgnore) {
+        originalConsoleError.apply(null, arguments)
+    }
+}
```
HendrikThePendric, line 24: This works as expected, but I'm not actually sure we want to add this. Ideally I would like to find a way to actually prevent these errors from showing in the log in the first place.

If we can't figure out how to do that, then perhaps we do want to keep this, but only when another condition is satisfied? For example:
```process.env.SUPPRESS_MISSING_COMPONENT_ERROR_LOG === true```

I don't like all this clutter to show during the test run, but I also worry that one day we will actually have a bad import/export somewhere and we don't see identify the problem due to this "fix"....

File: src/components/VisualizationPlugin/__tests__/VisualizationPlugin.spec.js [hunk 4]
```diff
@@ -170,7 +174,9 @@ describe('VisualizationPlugin', () => {
             // eslint-disable-next-line no-import-assign, import/namespace
             api.apiFetchAnalytics = jest.fn().mockRejectedValue('error')
 
+            const originalConsoleError = console.error
             await canvas()
+            console.error = originalConsoleError
 
             expect(defaultProps.onError).toHaveBeenCalled()
```
HendrikThePendric, line 181: Perhaps there is a more elegant way to achieve this, but I think this is probably acceptable here:
- We do in fact want an error to happen here
- We don't want to clutter the test output
- We want to assert that `defaultProps.onError` is called

HendrikThePendric: I had a go at testing this in a more conventional way:
```js
        it('calls onError callback when an exception is thrown', async () => {
            // eslint-disable-next-line no-import-assign, import/namespace
            api.apiFetchAnalytics = jest.fn().mockRejectedValue('error')

            expect(async () => {
                await canvas()
            }).rejects.toThrow()

            expect(defaultProps.onError).toHaveBeenCalled()
        })
 ```
 
But that doesn't work because the component does actually catch the error, so the test fails with the following message from Jest:
 ```
     expect(received).rejects.toThrow()

    Received promise resolved instead of rejected
    Resolved to value: undefined
 ```
 
 So I think that probably the solution I have above is correct. Even though it looks a bit hacky.

File: config/testSetup.js (hunk 3 above)
Comment (HendrikThePendric) [reply config/testSetup.js]: OK, found an actual solution to the problem in 4f97da5e99c87d63536f8c5d4d3f3c703179e8af

File: src/components/__tests__/App.spec.js [hunk 5]
```diff
@@ -10,6 +10,7 @@ import { UnconnectedApp as App } from '../App.js'
 import { Snackbar } from '../Snackbar/Snackbar.js'
 
 jest.mock('@dhis2/analytics', () => ({
+    ...jest.requireActual('@dhis2/analytics'),
```
HendrikThePendric, line 13: This is the fix for the component import errors.

src/components/DimensionsPanel/Dialogs/AddToLayoutButton/__tests__/AddToLayoutButton.spec.js
```diff
@@ -14,7 +14,7 @@ describe('The AddToLayoutButton component ', () => {
     }
 
     // TODO remove
-    console.log('getShallowAddToLayoutButton', getShallowAddToLayoutButton)
+    getShallowAddToLayoutButton
```
jenniferarnesen, line 17: This looks strange. Is this function being called?

jenniferarnesen (approved):

src/components/DimensionsPanel/Dialogs/AddToLayoutButton/__tests__/AddToLayoutButton.spec.js
jenniferarnesen: Seems like line 16 and 17 should just be removed.

HendrikThePendric: I agree and tried that at first.

But this has a knock-on effect with ESLint. Removing the line causes an error about the `getShallowAddToLayoutButton` function being defined but never used (`no-unused-vars`), and then removing that function caused a bunch of other ESLint errors, and before you know it the whole file is empty 🤣 .

I am not sufficiently familiar with this part of the codebase to quickly asses if removing so much "we should implement this later" code can/should be removed.

So the current change does look odd but is least intrusive and doesn't involve us having to disable the linter. So on balance I thought it was the least bad solution.

jenniferarnesen: Wouldn't it be better to use // eslint-disable-next-line instead?

HendrikThePendric: OK, I'll fix it like that

HendrikThePendric: After some more deliberation we decided to just delete that whole file.

PR #3220, opened 2024-09-27
feat: reduce flashing affect when hovering a highchart

jenniferarnesen:
Initially this PR uses a version of analytics that disables the "inactive" state on highcharts. There are also other settings we can look at changing and making configurable.

This change is actually for the dashboard slideshow, but trying it here in DV first for simplicity.


jenniferarnesen:
It was decided this is not useful

PR #3300, opened 2024-12-03
test: fix totals test by changing periods

janhenrikoverland:
Ref https://dhis2.slack.com/archives/G0108CD1421/p1733223595061669?thread_ts=1732787482.029189&cid=G0108CD1421

Fixes the totals test by relatively selecting the same months in the current year.

### DIFF
--- a/cypress/integration/options/totals.cy.js
+++ b/cypress/integration/options/totals.cy.js
@@ -58,66 +58,62 @@ describe('Options - Column totals', () => {
 
         clickOptionsModalHideButton()
     })
-    describe('Regression test for DHIS2-17297', () => {
-        it('does not crash', () => {
-            openContextMenu(DIMENSION_ID_DATA)
-            clickContextMenuMove(DIMENSION_ID_DATA, AXIS_ID_ROWS)
-            openContextMenu(DIMENSION_ID_PERIOD)
-            clickContextMenuMove(DIMENSION_ID_PERIOD, AXIS_ID_COLUMNS)
+    it('handles empty columns when the column totals option is enabled', () => {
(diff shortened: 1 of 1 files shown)


cypress/integration/options/totals.cy.js
```diff
@@ -47,18 +47,19 @@ import { TEST_CUSTOM_DIMENSIONS } from '../../utils/data.js'
 const AREA_DIMENSION = TEST_CUSTOM_DIMENSIONS.find((dim) => dim.name === 'Area')
 
 describe('Options - Column totals', () => {
-    describe('Regression test for DHIS2-17297', () => {
-        it('does not crash', () => {
-            goToStartPage()
-            changeVisType(visTypeDisplayNames[VIS_TYPE_PIVOT_TABLE])
+    beforeEach(() => {
```
jenniferarnesen, line 50: Could we give the tests proper names instead of "Regression test...". That title was only relevant the one time. Now the name should reflect the feature that is being tested.

cypress/integration/options/totals.cy.js
```diff
@@ -90,29 +90,33 @@ describe('Options - Column totals', () => {
...
+                'Monthly'
+            )
             clickDimensionModalUpdateButton()
 
             expectTableToBeVisible()
 
             // XXX is there a better way to address the total value cells?
-            expectTableValueCellToContainValue(12, 4)
-            expectTableValueCellToContainValue(13, 4)
+            expectTableValueCellToContainValue(8, 1)
```
jenniferarnesen, line 118: Maybe a FIXME or TODO instead of XXX?

jenniferarnesen (approved):

PR #3321, opened 2025-01-02
chore: use last year's periods

edoardo:
### Key features

1. use last year's period wherever possible

---

### Description

This is to avoid failing tests in the beginning of a new year when the new data is not added to the test DB yet.

### DIFF
--- a/cypress/integration/options/totals.cy.js
+++ b/cypress/integration/options/totals.cy.js
@@ -16,6 +16,7 @@ import {
     selectAllItemsByButton,
     selectDataElements,
     selectDataItems,
+    selectFixedPeriodYear,
     selectFixedPeriods,
     unselectAllItemsByButton,
 } from '../../elements/dimensionModal/index.js'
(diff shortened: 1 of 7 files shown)


cypress/integration/options/totals.cy.js
```diff
@@ -68,9 +69,10 @@ describe('Options - Column totals', () => {
         selectDataElements(['ART enrollment stage 1'])
         clickDimensionModalHideButton()
 
-        const year = new Date().getFullYear().toString()
+        const year = (new Date().getFullYear() - 1).toString()
```
jenniferarnesen, line 72: This is used so many times, perhaps it would make sense as a separate function `getPreviousYear` ?

edoardo: Agree, we do have that in LL.

edoardo: Done.

edoardo:
I converted to Draft since I realised the data is not the same for all the years, so some tests now fail because they expect a different value.
It's easy to fix the tests where there is actual data for both the current and the last year, but in some cases there is only data for the current year, so for those we would have to add the data to the DB dump.

PR #3325, opened 2025-01-27
chore: remove deprecated/removed object references (DHIS2-19062)

edoardo:
Implements [DHIS2-19062](https://dhis2.atlassian.net/browse/DHIS2-19062)

### Key features

1. remove deprecated `userCredentials` object references
2. remove deprecated `userAccesses`, `userGroupAccesses`, `externalAccess`, `publicAccess`

---

### Description

The `userCredentials` object is not returned inside the `user` object anymore.
The access objects are returned as part of `sharing`.
See: https://github.com/dhis2/dhis2-releases/blob/master/releases/deprecated-features.md#removed-features

[DHIS2-19062]: https://dhis2.atlassian.net/browse/DHIS2-19062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jenniferarnesen:
Confirmed that the following fields are no longer returned in v42:
* externalAccess
* publicAccess
* userAccesses
* userGroupAccesses

image


image

src/modules/fields/nestedFields.js
```diff
@@ -25,7 +24,6 @@ export const nestedFields = {
     columns: AXIS,
     rows: AXIS,
     filters: AXIS,
-    user: USER,
```
jenniferarnesen, line 28: You intended to remove the whole `user` object, right? Not just `userCredentials`

edoardo: I think since `user` is always returned with a set list of fields as you also mentioned above, it can be removed here where before we were only fetching `name` and `userCredentials`.
The change is then from `user[name,userCredentials[username]]` to `user`.

jenniferarnesen (approved):

PR #3335, opened 2025-02-05
chore: move analytics error handling in the plugin component

edoardo:
### Key features

1. Move analytics errors handling in the plugin component
2. Implement new custom error screen of org unit access error (E7120)

---

### Description

DV has a number of custom error screens where a user friendly error is shown to the user.
The `errorCode` returned in the analytics response is used to choose the error screen to show.

This error handling was done in the DV app components before, an `onError` callback was passed to the plugin component and used to pass the error back to the app.

Dashboard app also used the `onError` callback, but it didn't have all the custom error screen so a generic error was shown instead where the original error string from the analytics api was shown.
This error string is often obscure for the end user, and can also contain ids and other information intended more for admins or developers for debugging purposes.

When an error occured in a plugin, and the visualization was opened in the DV app, a different error was shown.

By moving the error handling in the plugin component, the same custom error screens are used in both the DV and dashboard apps.

---

### TODO

-   [ ] Cypress tests
-   [ ] Manual testing

---

### Screenshots

Before, no custom screen for E7120 in DV (already used in LL):
Screenshot 2025-02-05 at 15 52 36

After, new custom error for E7120 (same text as in LL):
Screenshot 2025-02-05 at 15 52 53

Same error screen used in dashboard app:
Screenshot 2025-02-05 at 15 52 59


src/components/VisualizationPlugin/VisualizationPlugin.js
```diff
@@ -145,6 +216,30 @@ export const VisualizationPlugin = ({
...
+                        visualization.type
+                    )
+
+                    result.responses.forEach((response) => {
+                        if (
+                            (response?.metaData?.dimensions || {})[
+                                DIMENSION_ID_DATA
+                            ]?.every(
+                                (dim) =>
+                                    response.metaData.items[dim]?.valueType ===
```
jenniferarnesen, line 231: On line 227 you check defined for both `response` and `metadata`. Any need to do that here as well?

src/components/Visualization/Visualization.js
```diff
@@ -212,25 +127,25 @@ export class UnconnectedVisualization extends Component {
             isLoading,
             onLoadingComplete,
         } = this.props
+
         const { renderId } = this.state
 
         return !visualization || error ? (
             <StartScreen />
```
edoardo, line 134: I'm considering removing the use of `VisualizationErrorInfo` from `StartScreen` and use it here directly when `error` is set.
I'm not 100% sure that `Visualization` is always rendered when there is an error set in Redux (`loader.loadingError`), if that is not the case then some errors won't be shown and most likely the app would crash.

edoardo, line 134: I'm considering removing the use of `VisualizationErrorInfo` from `StartScreen` and use it here directly when `error` is set.
I'm not 100% sure that `Visualization` is always rendered when there is an error set in Redux (`loader.loadingError`), if that is not the case then some errors won't be shown and most likely the app would crash.

HendrikThePendric: I agree that the fact that `StartScreen` can be an actual view containing "get started info" but could also show an error is pretty confusing. So removing it from here makes sense in any case. 

In theory, I am thinking that the `VisualizationPlugin` should be fully responsible for either showing an actual visualization, or a UI for informing the user that something went wrong while loading the visualization. And then that would mean that `VisualizationErrorInfo` is loaded in the plugin and nowhere else.

I actually think that it should be possible to do by looking at the code of the `Visualization` component (L133-156):
```js
return !visualization || error ? (
    <StartScreen />
) : (
    <Fragment>
        {isLoading && (
            <div style={styles.loadingCover}>
                <LoadingMask />
            </div>
        )}
        <VisualizationPlugin
            id={renderId}
            visualization={visualization}
            onChartGenerated={this.onChartGenerated}
            onLoadingComplete={onLoadingComplete}
            onDataSorted={this.onDataSorted}
            onResponsesReceived={this.onResponsesReceived}
            onDrill={this.onDrill}
            style={styles.chartCanvas}
            displayProperty={displayProperty}
        />
    </Fragment>
)
```
So currently we show the `StartScreen` when there is no visualization or when there is an error. Suppose we would only show the `StartScreen` if there is no `visualization`, and then we inform the `VisualizationPlugin` about the fact that there was a error by adding one more prop called `loadError` (we get the current `error` prop from the redux store using the `sGetLoadError` selector). 

In the `VisualizationPlugin` we add a `useEffect` hook that response to changes in the `loadError` prop and all we would need to do there is make some minor changes on L401-410:
```js
if (error) {
    return (
        <VisualizationErrorInfo
            error={formatError({
                error,
                visType: originalVisualization.type,
            })}
        />
    )
}
```
Here we could do something like this:
```js
const anyError = error || loadError
if (anyError) {
    return (
        <VisualizationErrorInfo
            error={formatError({
                anyError,
                visType: originalVisualization.type,
            })}
        />
    )
}
```


If needed, the `PluginWrapper` / `VisualizationPluginWrapper` could do the same as the `Visualization` component.

HendrikThePendric (commented):
I think this is an excellent idea and have added a comment to perhaps improve it even further.

src/components/Visualization/Visualization.js
HendrikThePendric: 🤔 I suppose instead of `loadError` we could have something like `externalError` or sth....

edoardo: I was thinking something as simple as render `VisualizationErrorInfo` in case of `error` and `StartScreen` for no visualization.

The reason is that there is at least one error which is not thrown by the plugin.
When you click Update, a layout validation happens, which can throw errors. This is only based on the UI selection done and happens before the plugin is rendered and the analytics request is made. In fact, this error should prevent the analytics request because it would be invalid anyway.
This error is currently rendered by `StartScreen` and goes the Redux route (`setError`, `loadError`)

With this PR the analytics errors are **not** going the Redux route, since the plugin doesn't use Redux and the handling is now internal to `VisualizationPlugin`.

To summarise we could:
1) render `VisualizationErrorInfo` in `Visualization` in case of `error`
2) remove the use of `VisualizationErrorInfo` and related error getter from `StartScreen`
3) move `VisualizationErrorInfo` outside of the Visualization folder since it's also used by the plugin directly.

The plugin wrapper components don't need any error handling because they cannot be connected to Redux and we assume that the `visualization` object passed is a valid one that can generate proper analytics requests.
Of course the analytics request itself can error, ie. no data, no access... but those are all coming from the analytics response and are now handled in the plugin.

HendrikThePendric (approved):

PR #3362, opened 2025-03-12
fix: fix regression in Not found visualization error (DHIS2-19187)

edoardo:
Implements [DHIS2-19187](https://dhis2.atlassian.net/browse/DHIS2-19187)

---

### Key features

1. fix regression with Not found error not showing up

---

### Description

If a visualization is not found due to using an outdated link or a wrong id in the URL, a custom message should be shown informing about it.
Since v101.0.0 this stopped working.
The culprit was a change introduced when refactoring the error handling.

---

### TODO

- [ ] Dashboard tested

---

### Known issues

- [x] an error different than a 404 does not cause the app to redirect to `/` which has a problem when using the Update button. This was fixed in #3056 but only for the 404 case. Should the redirect happen for other errors as well? A 405 error is easily triggered by an invalid id (ie. remove 1 or more characters from the id in the URL). **NB: this is not a regression introduced by this PR, it's an existing issue that we might want to address in a separate PR.**

---

### Screenshots

Before:
Screenshot 2025-03-12 at 11 13 55

After:
Screenshot 2025-03-12 at 11 13 27

[DHIS2-19187]: https://dhis2.atlassian.net/browse/DHIS2-19187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -114,14 +114,12 @@ export const tDoLoadVisualization =
             if (errorResponse?.details?.httpStatusCode === 404) {
                 error = new VisualizationNotFoundError()
                 history.push('/')
-            } else if (errorResponse?.message) {
-                error = errorResponse.message
-            } else {
+            } else if (!errorResponse?.message) {
                 error = new GenericServerError()
             }
             dispatch(clearAll(error))
 
-            logError('tDoLoadVisualization', error)
+            logError('tDoLoadVisualization', error.description || error.message)
         }
     }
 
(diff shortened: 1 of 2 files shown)


HendrikThePendric (approved):
After receiving some additional background information from @edoardo on Slack, I think these changes make sense.

PR #3375, opened 2025-04-01
fix:  restore navigation between visualizations using the browser address bar [DHIS2-19387]

KaiVandivier:
Implements [DHIS2-19387](https://dhis2.atlassian.net/browse/DHIS2-19387)

Manual URL changes trigger location updates with the key `default`, which then results in the `history` skipping its handler because it's the same as the key as the previous updates

It's still important to check they `key` to avoid triggering the handler multiple times for the same location update, but the pathname and search can be checked to see if the rest of the handler should run

https://github.com/user-attachments/assets/b23326ad-c723-4798-94d4-cee609c94036

[DHIS2-19387]: https://dhis2.atlassian.net/browse/DHIS2-19387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -35,7 +35,7 @@ import './App.css'
 import './scrollbar.css'
 
 // Used to avoid repeating `history` listener calls -- see below
-let lastLocationKey
+let lastLocation
 
 export class UnconnectedApp extends Component {
     unlisten = null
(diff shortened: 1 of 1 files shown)


HendrikThePendric (approved):
LGTM, thanks @KaiVandivier

HendrikThePendric (changes requested):
I prematurely approved this PR but do not want to see it merged until I have verified this actually solves the E2E failure in push-analytics. The only way to "remove approval" in GitHub seems to be to request changes, so that's what I am doing here, even though I think the implementation looks good.

HendrikThePendric (approved):
I have created a build-bundle from this branch and installed it to the local DHIS2 Core Instance that I use for testing push-analytics. I've been able to confirm that this navigation fix resolves the issue there.

So I am re-approving the PR.

PR #3377, opened 2025-04-08
fix: rename visualization fails in certain cases [DHIS2-19433]

jenniferarnesen:
Fixes [DHIS2-19433](https://dhis2.atlassian.net/browse/DHIS2-19433)

### Description

Previously, renaming was done in the analytics FileMenu->RenameDialog using api `PATCH`. However, several backend bugs have been found with the use of `PATCH`. This commit is a workaround. It uses a `PUT` (same as the Save functionality) and saves the entire original visualization with the updated name/description, but without any other unsaved changes.

Note that this implementation also fixes an unreported bug where users cannot delete a description.

### TODO

- [ ] Dashboard tested N/A
- [x] Tests added (Cypress and/or Jest)
- [ ] Docs added N/A
- [x] d2-ci dependency replaced
- [x] Testing sign-off (checked by the tester, not developer)

Requires https://github.com/dhis2/analytics/pull/1771

[DHIS2-19433]: https://dhis2.atlassian.net/browse/DHIS2-19433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/actions/__tests__/index.spec.js
```diff
@@ -246,22 +250,52 @@ describe('index', () => {
...
 
         const extraParams = {
-            name: 'rename-test',
-            description: 'Rename test',
+            name: 'Renamed name',
+            description: 'Renamed description',
         }
 
+        // eslint-disable-next-line no-import-assign, import/namespace
+        api.apiSaveVisualization = jest.fn(() => {
```
edoardo, line 280: Is this one necessary? It's overridden inside each `it` block.

edoardo (approved):

PR #3382, opened 2025-04-15
chore: add packageManager field to package.json

HendrikThePendric:
This adds a new field to the `package.json` file called `packageManager`?

This is for devs using NodeJS' [corepack](https://github.com/nodejs/corepack). After calling `corepack enable` and uninstalling the global version of `yarn` as per the corepack docs, whenever you switch to a new repo, the package manager specified in the `packageManager` will be used. So practically speaking for DHIS2 this allows us to use different versions of `yarn` in different repos.


HendrikThePendric:
As agreed during the team meeting: these changes to `package.json` can be sneaked in alongside other PRs. Closing this PR now.

PR #3384, opened 2025-04-16
fix: fetch latest vis before renaming so not to overwrite others changes

jenniferarnesen:
What if we fetch the latest before renaming? Potential risks include what will happen in the app when we store the latest vis in redux after renaming.


jenniferarnesen:
not now

PR #3386, opened 2025-04-23
fix: fetch latest visualization before renaming so not to overwrite newer changes

jenniferarnesen:
Fixes [DHIS2-19506](https://dhis2.atlassian.net/browse/DHIS2-19506)

### Description

Fetch the visualization and apply name and description to that object, which is then used in the rename PUT request. In the app, only name and description are updated, which means the app's copy of the visualization could still be out of date. But that's a different problem that needs a different solution.

---

### Quality checklist

- [ ] Dashboard tested N/A
- [ ] Tests added (Cypress and/or Jest)
- [ ] Docs added N/A
- [ ] Global shell tested N/A
- [ ] d2-ci dependency replaced N/A

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -194,10 +194,13 @@ export const tDoRenameVisualization =
         }
 
         try {
+            const { visualization } = await apiFetchVisualization(
+                engine,
+                sGetVisualization(getState()).id
+            )
+
             const visToSave = await preparePayloadForSave({
-                visualization: getSaveableVisualization(
-                    sGetVisualization(getState())
-                ),
+                visualization: getSaveableVisualization(visualization),
                 name,
                 description,
                 engine,


edoardo (approved):

src/actions/index.js
```diff
@@ -152,58 +153,59 @@ export const clearAll =
...
 
-                updatedVisualization.description = updatedCurrent.description =
-                    description
+            const visualization = sGetVisualization(state)
+            const current = sGetCurrent(state)
 
-                dispatch(
-                    fromVisualization.acSetVisualization(updatedVisualization)
-                )
+            const updatedVisualization = { ...visualization, ...updatedVis }
```
edoardo, line 164: Here we merge the "old" visualization with the new name/description.
Is this correct?
`visualization` usually reflects what is stored in the DB, and although it can get out of sync, with the changes in the PR we do fetch the "latest" from the DB, so I'm wondering if this merge should use the "latest" instead of whatever we had in Redux.

jenniferarnesen: Yes I intentionally only update the name/desc in the redux `visualization` prop. This was to limit the scope of this PR. I wasn't sure what side effects there would be to updating the whole object.

edoardo (approved):

cypress/integration/saveAndRename.cy.js
```diff
@@ -110,12 +110,24 @@ describe('rename', () => {
...
 
         // rename the AO, changing name only
         renameVisualization(UPDATED_AO_NAME)
 
+        cy.wait('@get-visualization')
         cy.wait('@put-rename')
+        cy.wait('@get-visualization')
+
+        // Assert that there were exactly 2 GET requests
+        cy.get('@get-visualization.all').should('have.length', 2)
```
edoardo, line 130: Is this `.all` an `alias` feature?

cypress/integration/saveAndRename.cy.js
```diff
@@ -318,15 +330,61 @@ describe('saving an AO', () => {
...
+        cy.contains('About this visualization').should('not.exist')
+
+        // check first GET request fetches subscribers
+        cy.intercept(
+            'GET',
+            /\/api\/\d+\/visualizations\/\w+\?fields=subscribers/
+        ).as('get-subscribers')
+
+        // check the save request contains subscribers
+        cy.intercept('PUT', /\/api\/\d+\/visualizations\/\w+(\?.*)?/, (req) => {
```
edoardo, line 359: Is the last part of the regexp needed since it's optional anyway?

cypress/integration/saveAndRename.cy.js
```diff
@@ -110,12 +110,24 @@ describe('rename', () => {
         expectAOTitleToBeValue(AO_NAME)
         expectVisualizationToBeVisible(VIS_TYPE_PIVOT_TABLE)
 
-        cy.intercept('PUT', '**/api/*/visualizations/*').as('put-rename')
+        cy.intercept('GET', /\/api\/\d+\/visualizations\/\w+\?fields=.*/).as(
+            'get-visualization'
+        )
+
+        cy.intercept('PUT', /\/api\/\d+\/visualizations\/\w+(\?.*)?/, (req) => {
```
edoardo, line 117: The last part the regexp `(\?.*)?` seems to be unnecessary

cypress/integration/saveAndRename.cy.js
```diff
@@ -318,15 +330,61 @@ describe('saving an AO', () => {
...
         expectVisualizationToBeVisible(TEST_VIS_TYPE)
 
+        // check user is still subscribed
+        cy.getBySel('dhis2-analytics-interpretationsanddetailstoggler').click()
+        cy.get('button').contains('Unsubscribe').should('be.visible')
+        cy.getBySel('dhis2-analytics-interpretationsanddetailstoggler').click()
+        cy.contains('About this visualization').should('not.exist')
+
         // saves AO using "Save As"
+        cy.intercept('POST', /\/api\/\d+\/visualizations(\?.*)?/, (req) => {
```
edoardo, line 377: Same thing here.

src/modules/fields/index.js
```diff
@@ -6,8 +6,8 @@ import {
 } from './baseFields.js'
 import { extendFields } from './nestedFields.js'
 
-export const getFieldsStringByType = (type) =>
-    getAllFieldObjectsByType(type)
+export const getFieldsStringByType = (type, withSubscribers) =>
```
edoardo, line 9: Minor, I would use an object for the "options" here:
```suggestion
export const getFieldsStringByType = (type, { withSubscribers }) =>
```

edoardo (commented):
Left some comments about the regexp used in the intercept.

edoardo (approved):

PR #3396, opened 2025-05-19
fix: avoid crash when switching vis type due to unsupported options (DHIS2-19611)

edoardo:
Implements [DHIS2-19611](https://dhis2.atlassian.net/browse/DHIS2-19611)
Fixes [DHIS2-19539](https://dhis2.atlassian.net/browse/DHIS2-19539) via analytics

### Description

When switching to a different visualization type, depending on the configured options, crashes might happen.
Not all options are relevant for all visualization types, but all the options are saved, loaded and configured (with defaults) in the app's UI.
Before we passed the whole set of options to the plugin regardless of the visualization type, and this caused crashes with different type / set of options combinations.
Now the object passed to the plugin is filtered to remove all the irrelevant options.

---

### Quality checklist

- [x] Dashboard tested
- [x] Cypress and/or Jest tests added/updated
- [ ] Docs added N/A
- [x] d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/1780)
- [x] Tester approved (JJA)

---

### Screenshots

_supporting images_

[DHIS2-19611]: https://dhis2.atlassian.net/browse/DHIS2-19611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-19539]: https://dhis2.atlassian.net/browse/DHIS2-19539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reducers/current.js
```diff
@@ -92,17 +92,17 @@ export const sGetCurrentFromUi = (state) => {
 
     switch (ui.type) {
         case VIS_TYPE_PIE:
-            return getPieCurrentFromUi(state, ui)
+            return getPieCurrentFromUi(state.current, ui)
```
edoardo, line 95: I found out we were passing the whole Redux state here instead of just the `current` slice like in other parts of the codebase.
This required to adjust the layout validation for Scatter where `ui` is needed and was read from this "current" object.

src/components/Visualization/Visualization.js
```diff
@@ -215,6 +218,17 @@ export const Visualization = (props) => {
     )
     const { setChart } = useChartContext()
 
+    const visualization = useMemo(() => {
+        const currentClone = cloneDeep(current)
+
+        currentClone &&
+            new Set(Object.keys(options))
```
edoardo, line 225: This removes all the options that are not relevant for the visualization type.
Because all the options are at the root level of the object it's hard to filter the options otherwise without affecting other non-option keys.

src/modules/options/defaultConfig.js
```diff
@@ -53,3 +53,50 @@ export default ({
...
+    isStacked,
+    supportsLegends,
+    isColumnBased,
+}) => {
+    const options = [
+        'cumulativeValues',
+        'hideEmptyRowItems',
+        'sortOrder',
+        'skipRounding',
+
```
jenniferarnesen, line 67: Are the empty lines added to separate between options going on different tabs in the ui? Maybe a comment instead of empty line would be more informative.

src/modules/options/outlierTableConfig.js
```diff
@@ -38,3 +38,14 @@ export default () => [
         },
     ]),
 ]
+
+export const outlierTableOptionNames = () => [
+    'skipRounding',
```
jenniferarnesen, line 43: Use constants instead of repeating the strings in each config.

src/reducers/current.js
jenniferarnesen: ok.

jenniferarnesen (changes requested):
Setting to request changes while the comments are addressed.

src/components/Visualization/Visualization.js
edoardo: Extracted the logic into a function and added unit tests.

src/modules/options/pivotTableConfig.js
```diff
@@ -126,3 +159,44 @@ export default ({
         ],
     },
 ]
+
+export const pivotTableOptionNames = () => [
+    // Data tab
+    OPTION_CUMULATIVE_VALUES,
+    OPTION_SHOW_DIMENSION_LABELS,
+    OPTION_SKIP_ROUNDING,
+    'colSubTotals',
```
jenniferarnesen, line 168: OPTION_COL_SUB_TOTALS?

edoardo: Fixed.

jenniferarnesen (approved):

jenniferarnesen (dismissed):

jenniferarnesen (approved):

PR #3402, opened 2025-05-22
fix: make jspdf and svg2pdf deps available for LAN only setups

HendrikThePendric:
Implements [DHIS2-19646](https://dhis2.atlassian.net/browse/DHIS2-19646)

### Description

Sometimes a client can be connected to a local DHIS2 server, but not to the internet. I refer to this as a LAN-only setup. In these cases the DHIS2 Connection status will indicate that the client is connected and the "Export to PDF" menu option is available. However, prior to the changes in this PR, the export would fail silently because the app would attempt to download the PDF conversion libraries from the code.highcharts.com domain.

So we need to host these files with the app and I have chosen to implement this as follows:
- These files are actually static assets that are loaded on-demand. Because of this I have excluded them from the pre-cache, just like the font (`.ttf`) files.
- Manually keeping these files up-to-date would be very inconvenient, so I wanted to add them to the app's dependencies which ensures that `dependabot` raises pull requests to keep them up to date. Since these are not regular dependencies that get imported by the app, I decided that adding them as `devDependencies` would be more appropriate.
- Finally, to ensure that the files are indeed updated regularly, I added a `postinstall` script to `package.json`. This is a built-in feature of [yarn 1.22.22](https://classic.yarnpkg.com/en/docs/package-json#toc-scripts), once `install` completes `postinstall` runs. This script copies the files to the `./public/vendor/` dir.

The above is perhaps slightly unconventional, but since we don't have direct access to a webpack config file, I actually think this is the most simple and convenient solution. It will break once one of the dependencies decides to rename/move/remove the relevant files, but when that happens it will be pretty obvious and should be easy to address.

---

### Quality checklist

- [x] N/A Cypress and/or Jest tests added/updated
- [x] N/A Docs added
- [x] N/A d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/XXX)
- [x] Tester approved (Edoardo)
- [x] N/A Dashboard tested

---

### Expected outcome

- When doing a PDF export these dependencies are fetched from `./vendor/` and no longer from `code.highcharts.com`
- The dependencies are not being pre-cached
- Exporting to PDF still works as expected

---

### Screenshots

Below are some screenshots from the network tab that show that all files required by the PDF exporting process are now downloaded from the app's public dir, which means the app will work in a LAN-only setup:

Screenshot 2025-05-22 at 17 02 02
Screenshot 2025-05-22 at 17 02 23
Screenshot 2025-05-22 at 17 02 53
Screenshot 2025-05-22 at 17 03 10
Screenshot 2025-05-22 at 17 03 28
Screenshot 2025-05-22 at 17 03 41

[DHIS2-19646]: https://dhis2.atlassian.net/browse/DHIS2-19646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edoardo (commented):
Tested on the Netlify link.
Changes look good, clean solution.

PR #3466, opened 2026-04-21
fix: select option widths

cooper-joe:
Fixes [DHIS2-14363](https://dhis2.atlassian.net/browse/DHIS2-14363)

### Description

This PR allows adjusts the `SelectBaseOption` to accept an `inputWidth`. This allows options that anticipate longer labels to set a more generous width. These widths are fixed, so will overflow their containers if there is insufficient space. This is only an issue on small screens and this app is not optimised for mobile, so I think this is a worthy tradeoff for simplicity. 

Changes made to the following selects:
- Aggregation type: increased to `540px`. This makes it single line for French, too.
- Legend selection: increased to `540px` as it is user-generated content (this select didn't actually use `SelectBaseOption`, but I've updated it here as I consider it in scope).

---

### Quality checklist

- [x] Dashboard tested _N/A_
- [x] Cypress and/or Jest tests added/updated _N/A_
- [x] Docs added _N/A_
- [x] Tester approved (@karolinelien )

---

### Screenshots

**Before:**

image

**After:**

SCR-20260421-dw3

SCR-20260421-dvd

[DHIS2-14363]: https://dhis2.atlassian.net/browse/DHIS2-14363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


karolinelien:
Looks great from a user facing perspective!

src/components/VisualizationOptions/Options/LegendSet.jsx
```diff
@@ -36,7 +36,7 @@ const LegendSetSelect = ({
         label={i18n.t('Legend')}
         disabled={disabled}
         selected={value?.id}
-        inputWidth="280px"
+        inputWidth="540px"
```
janhenrikoverland, line 39: @cooper-joe  Should 540px be a constant/standard for wider components?

cooper-joe: It could be. Could be:
```
280: default
540: wide
100%: fullwidth
```

But I'm not sure we need the standardisation. When an input needs an adjusted width for some reason, I feel like it's just as likely it'll need something other than the standards, so we'd end up with overrides anyway.

Not a strong feeling though, so happy to standardise if you'd prefer.

janhenrikoverland: Not important

janhenrikoverland (approved):

PR #3470, opened 2026-06-09
chore: use dhis2/action-commit-signing

janhenrikoverland:
Ref https://github.com/dhis2/action-commit-signing/blob/master/README.md


radnov (approved):
LGTM 👍

